PR #155, opened 2011-09-18
Documentation for Pointer, Type, Type::Builtin, Type::Mapped, NativeType, StrPtrConverter

sdaubert:
Documentation for FFI::Pointer, FFI::Type, FFI::Type::Builtin, FFI::Type::Mapped, FFI::NativeType, FFI::StrPtrConverter.
Minor correction for FFI::AbstractMemory#put_bytes (old doc seems to be for #get_bytes).

I have some problems with documentation due to RDoc which uses variable names to guess parent class from rb_define_class or rb_define_class_under. As varibale are named "NameClass" (where "Name" is the class name), RDoc guesses parent class is NameClass instead of Name.
A typical example is FFI::Pointer. As its parent is guessed AbstractMemoryClass, no link to browse to AbstractMemory documentation page appears on Pointer page.
I think this is harmful for documentation quality and for easiness to use.

Here is my question : may I change name variables in my future work by using a ruby-like naming style (rbffi_cName instead of rbffi_NameClass by example).

### DIFF
--- a/ext/ffi_c/AbstractMemory.c
+++ b/ext/ffi_c/AbstractMemory.c
@@ -458,12 +458,12 @@ memory_get_bytes(VALUE self, VALUE offset, VALUE length)
 
 /*
  * call-seq: memory.put_bytes(offset, str, index=0, length=nil)
- * Return string contained in memory.
+ * Put a string in memory.
  * @param [Numeric] offset point in buffer to start from
  * @param [String] str string to put to memory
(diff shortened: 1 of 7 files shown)


steakknife:
Lol oops I meant closes [libvips/#155](https://github.com/libvips/ruby-vips/issues/155). Nothing to see here, move along. :P

PR #315, opened 2013-12-12
Use Ruby implementation for `which`

davispuh:
In Windows there's no `/dev/null` and it causes error "The System cannot find the path specified". Also in Windows there's no native `which` and it works only because it's included in DevKit.

This Ruby implementation should correctly work for all platforms.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -69,7 +69,18 @@ OS = case RbConfig::CONFIG['host_os'].downcase
     RbConfig::CONFIG['host_os'].downcase
   end
 
-GMAKE = system('which gmake >/dev/null') && 'gmake' || 'make'
+def which(name)
+  exts = ENV['PATHEXT'] ? ENV['PATHEXT'].split(';') : ['']
+  ENV['PATH'].split(File::PATH_SEPARATOR).each do |path|
+    exts.each do |ext|
+      app = File.join(path, name+ext)
+      return app if File.executable? app
+    end
+  end
+  nil
+end
+
+GMAKE = which('gmake').nil? ? 'make' : 'gmake'
 
 LIBTEST = "build/libtest.#{LIBEXT}"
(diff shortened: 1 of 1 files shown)


djberg96:
Please merge.

tduehr:
Why iterate over the path and extensions rather than use a glob?

djberg96:
@tduehr wouldn't work on Windows with their file extensions, e.g. notepad.exe, not notepad.

tduehr:
`Dir.glob("*.{exe,bat}")` works for me in 2.2.2 (compiled w/ VC) under win8.1. `C:\Ruby\bin`

`Dir.glob("**/*{#{ENV["PATHEXT"].tr(';',',')}}")` works in the directory above it `C:\Ruby`

djberg96:
Hm, for some reason I thought that approach had some issue, but perhaps I'm mistaken. Anyway, it's not like we need to hyper optimize for this little bit of functionality. I'm happy with whatever works. :)

davispuh:
I'm don't really understand why glob would be preferable? I'm not sure if there would be any performance benefits.
Here's implementation using Dir.glob and it seems to be more complicated...
```ruby
def which(name)
  exts = ENV['PATHEXT'] ? '{' + ENV['PATHEXT'].split(';').join(',') + '}' : ''
  paths = '{' + ENV['PATH'].split(File::PATH_SEPARATOR).join(',').gsub('\\', '/') + '}'
  Dir.glob(paths + '/' + name + exts).each do |app|
    return app if File.executable?(app)
  end
  nil
end
```

In PATH there can be multiple executables with same name and some might not even be executable so we've to check all in correct order.

tduehr:
I was wondering if there was a specific reason is all... Then i got to way over optimizing. I'll pull this when I get a chance to write something to use which normally then failover to this.

PR #357, opened 2014-07-28
Add version display

domgetter:
Added code and spec so that FFI::VERSION now displays the current ffi version.  Let me know if you want anything changed, different kind/style of test, or if you don't want the feature at all.

- Dominic


domgetter:
Oh, was there a reason you didn't want FFI::VERSION to work?

Issue #360, opened 2014-08-07
ffi won't compile on Solaris 11.x


efrancis:
When trying to install ffi on Solaris 11 (tested 11.1 and 11.2), you get prompted that you need to install developer tools first (even though everything is installed).

If you comment out the line that was added for Solaris 10 in ext/ffi_c/extconf.rb, then ffi installs successfully:
     # $CFLAGS << " -std=c99" if RbConfig::CONFIG['host_os'] =~ /solaris/

bahamat:
There are two pull requests open to address this issue. #382 and #391. Can we get consensus on which one should be merged?

tduehr:
Neither. They don't address the whole thing. I'll get this ironed out in the morning.

tduehr:
Give #399 a try. It should work for both solaris 10 and 11. I've only tested it on 11 so far.

efrancis:
Tested on Solaris 11.2 and works perfectly now.

bahamas10:
#382, #391 and #399 all tested and working on SmartOS.

PR #382, opened 2014-10-27
#360:  Use correct compiler flag when using Solaris Studio.

orderthruchaos:
This will also allow compiler flag to be set for older/newer versions of Solaris.  Admittedly, it is a bit of a preemptive strike for later versions.

This should help with issue #360.

PR #391, opened 2014-11-26
#360: Fix solaris 10 match so it doesn't match solaris 11

face:
'solaris2.11'  =~ /solaris(!?2.11)/ will return 0, which in ruby is true, not false so std=c99 is still showing up on solaris 11.


bahamat:
This pull request overlaps with #382, fixing the same issue. Can we get consensus on which one should be merged?

tduehr:
Please try #399 as a fix for this.

sax:
I will try this fix on SmartOS as soon as I can.

bahamas10:
Tested and working on SmartOS

bixu:
+1 to fixing this - what can I do to help?

smartb-pair:
bump 😃

tduehr:
Would pulling this flag from RbConfig work? I don't have a Solaris box to test this properly yet. What do I need to install in a VM to replicate your environment?

PR #393, opened 2014-11-26
Memory alignment fix for VS2013 x64

rafalbigaj:
On Windows x64 compilation with VisualStudio 2013 any call to ffi function ends with segmentation fault.
It's caused by invalid memory alignment. The used mask 0x7UL is 32 bit on this platform, 0x7ULL is required.


cfis:
Any chance this can be applied? I have just run into the same issue, and the patch correctly fixes it.

PR #405, opened 2015-01-24
Fixed the spec for the FFI:Function's :blocking parameter

DavidEGrayson:
Quick intro: FFI:Function can wrap a C function, and if the `:blocking` parameter is set to true then it should release MRI's global interpreter lock (GIL) when you call the function, allowing multiple blocking functions to run concurrently.

You can tell the old spec was bad because you could remove `:blocking => true` from the spec and it would still pass.  I fixed it so that it actually tests that multiple blocking functions will run concurrently.  If they are not running concurrently, the spec is guaranteed to take at least 10 seconds to finish and it will therefore fail.

### DIFF
--- a/spec/ffi/function_spec.rb
+++ b/spec/ffi/function_spec.rb
@@ -62,16 +62,16 @@ describe FFI::Function do
     expect(foo.add(10, 10)).to eq(20)
   end
 
-  it 'can wrap a blocking function' do
+  it 'can run multiple blocking functions concurrently' do
     fp = FFI::Function.new(:void, [ :int ], @libtest.find_function('testBlocking'), :blocking => true)
+    time = Time.now
     threads = 10.times.map do |x|
       Thread.new do
-        time = Time.now
-        fp.call(2)
-        expect(Time.now - time).to be > 2
+        fp.call(1)
       end
     end
     threads.each { |t| t.join }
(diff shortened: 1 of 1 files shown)


tduehr:
This is the same logic that the current test replaced. It fails regularly on travis when the testing VM runs slow.

What we need is a test for concurrency that doesn't rely on timing.

DavidEGrayson:
Ok, I found the last time the test had been changed:

https://github.com/ffi/ffi/commit/c66961af7fcc44121a7d060734d300afd3120419

The old version of the test before that commit was not very good since it creates a new thread and asserts that the thread can start within 1 second, so if the machine is doing something else for one second at the wrong time then it could fail.  Also, it is not guaranteed to catch bugs that cause the function to run non-concurrently.

The propose change in this pull request just asserts that all the child threads can finish within 9 seconds, so I think it should be less likely to fail since the time frame is longer.

Still, I agree that we can do better.

If we could make a variable that can be accessed atomically from multiple threads, it would be easy to verify that a function can run concurrently.  Do we have access to any threading libraries that might have a feature like that?  I would be tempted to use [GCC's atomic builtins](https://gcc.gnu.org/onlinedocs/gcc-4.1.2/gcc/Atomic-Builtins.html) but does the FFI gem support running tests on compilers other than GCC?

PR #406, opened 2015-01-25
Fixed the spec for the FFI::Function's :blocking parameter (try #2)

DavidEGrayson:
This is my second attempt to fix the spec for the FFI:Function blocking parameter.  For more information, see the first attempt:

https://github.com/ffi/ffi/pull/405

This version passes a message from one thread to another in a way that requires two different threads to be running the C code concurrently.  It is not affected by the speed of the computer it is running on.  (On the WIN32 platform, the test will not test anything, but it should at least pass.  It would be nice to get pthreads working on Windows though.)

### DIFF
--- a/spec/ffi/fixtures/FunctionTest.c
+++ b/spec/ffi/fixtures/FunctionTest.c
@@ -24,9 +24,46 @@ int testFunctionAdd(int a, int b, int (*f)(int, int))
     return f(a, b);
 };
 
-void testBlocking(int seconds) {
-    sleep(seconds);
-};
+// The first thread that starts running this function will set the
+// flag variable to 1 and then wait for another thread to set it back
+// to 0, which confirms that two copies of the function were running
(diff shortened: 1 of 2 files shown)


tduehr:
This seems to hang when the test fails. `:blocking => false` causes the test to hang.

Is there a way to get a failure condition out of the test here?

DavidEGrayson:
Yeah, I think a timeout could be implemented in the `while(flag)` loop.  You could use that to tell when 30 seconds have passed, and then break out of the loop.

tduehr:
Any progress on this?

DavidEGrayson:
Right now the FFI gem still has a test in it that doesn't test anything.  I would think this pull request should be left open as a reminder that someone, some day, should fix that eventually.  Or perhaps putting a "TODO" note in the spec itself would be better.

tduehr:
... That's why I'd left it open originally in the first place....

tduehr:
Fixed by #451

Issue #413, opened 2015-03-02
Fails to compile on RHEL7 little endian ppc64le


kaustubh-d:
Fails to compile on RHEL7 little endian ppc64le. Has anybody tried and got it working on this platform?

First failure is 
Configuring libffi
/home/jenkins/kaustubh/thirdparty/kdffi/ffi/ext/ffi_c/libffi/config.guess: unable to guess system type
...
This platform guess is overcome by using the latest config.guess
http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD

Next issue is:
libtool: link: gcc -shared  -fPIC -DPIC  src/.libs/debug.o src/.libs/prep_cif.o src/.libs/types.o src/.libs/raw_api.o src/.libs/java_raw_api.o src/.libs/closures.o src/powerpc/.libs/ffi.o src/powerpc/.libs/sysv.o src/powerpc/.libs/ppc_closure.o src/powerpc/.libs/linux64.o src/powerpc/.libs/linux64_closure.o      -Wl,-soname -Wl,libffi.so.5 -o .libs/libffi.so.5.0.10
/usr/bin/ld: src/powerpc/.libs/linux64.o: ABI version 1 is not compatible with ABI version 2 output
/

Again if I pull in latest libffi source code, am able to get it compile with some tweaks, but ffi gem tests would fails for ./spec/ffi/variadic_spec.rb, example
  595) Function with variadic arguments call(dfj, double,9.87654321,float,1.23456789,long_long,2211230779876286898)
     Failure/Error: expect(p.get_float64(off)).to eq(v)

```
   expected: 9.87654321
        got: 3.4766470625842e-310
```

 Seems extension code will need some porting to work with latest libffi source code.

Same issues for Ubuntu 14 LE on ppc64le

kaustubh-d:
Any inputs on this?

tduehr:
These platforms are supported on a "this *should* work" basis.

That said, we're working on getting a process together for building VMs for all the supported platforms to enable actual testing across the board.

If you update libffi (either in the repo or by installing libffi-devel) does that correct the problem?

Do the ffi tests pass under jruby?

mattray:
I talked with IBM about the issue and they pointed to this commit back in August that was merged to master:
https://github.com/ffi/ffi/commit/30f155d05cac28d82baf0283404b675222616f9d

[...]

Issue #201, opened 2012-04-18
rake specs failure on Windows


djberg96:
Hi,

ffi (latest from git repo) using MSVC++ 10
Windows 7

<pre>
c:\Users\djberge\Repositories\ffi>rake specs
The system cannot find the path specified.
[warn] YARD unavailable
mkdir -p build/i386-mswin32_100/ffi_c/1.9.3
cd build/i386-mswin32_100/ffi_c/1.9.3
c:/usr/bin/ruby.exe -I. ../../../../ext/ffi_c/extconf.rb
checking for ffi.h... no
checking for ffi.h in /usr/local/include... no
checking for rb_thread_blocking_region()... yes
checking for ruby_native_thread_p()... yes
checking for rb_thread_call_with_gvl()... yes
creating extconf.h
creating Makefile
cd -
cd build/i386-mswin32_100/ffi_c/1.9.3
nmake

Microsoft (R) Program Maintenance Utility Version 10.00.30319.01
Copyright (C) Microsoft Corporation.  All rights reserved.

generating ffi_c-i386-mswin32_100.def

Microsoft (R) Program Maintenance Utility Version 10.00.30319.01
Copyright (C) Microsoft Corporation.  All rights reserved.

NMAKE : fatal error U1052: file 'Makefile.vc' not found
Stop.
NMAKE : fatal error U1077: 'cd' : return code '0x2'
Stop.
rake aborted!
Command failed with status (2): [nmake...]

Tasks: TOP => specs => compile => compile:i386-mswin32_100 => compile:ffi_c:i386-mswin32_100 => copy:ffi_c:i386-mswin32_
100:1.9.3 => build/i386-mswin32_100/ffi_c/1.9.3/ffi_c.so
(See full trace by running task with --trace)
</pre>

djberg96:
I still see this on master with VC 2013.

tduehr:
`rake specs` doesn't work anyway... does `rake spec` work with VC?

djberg96:
@tduehr Same issue with rake spec.

tduehr:
does `rake build` work?

djberg96:
@tduehr Yes, it works.

tduehr:
can you supply a fresh full trace for this?

djberg96:
The whole shebang:

    PS C:\Users\djberge\Repositories\ffi> git pull
    Current branch master is up to date.

    PS C:\Users\djberge\Repositories\ffi> ruby -v
[...]

PR #424, opened 2015-03-18
Update embedded libffi to 3.2.1

kaustubh-d:
Update embedded libffi to 3.2.1 so that gem can be successfully built and installed on ppc64le.

This version of libffi adds support for PowerPC 64 Little and Big Endian archs. See #413


tduehr:
There's a bit more to upgrading libffi than just copying it in... I'll have a branch w/ it updated for some additional testing in a bit.

tduehr:
@djberg96 Can you test the libffi-3.2.1 branch on windows?

djberg96:
@tduehr I'm still having this issue: https://github.com/ffi/ffi/issues/201

kaustubh-d:
@tduehr - no problem! realized that when trying on new system. One issue I see is the extension build should trigger the internal libffi configure and then use the generated ffi.h to set the extension build flags (extconf.h) from here https://github.com/ffi/ffi/blob/master/ext/ffi_c/extconf.rb#L28-L31.

PR #435, opened 2015-05-13
Enable and fix warnings (esp. in Enum#native_type)

mvz:
This pull request fixes a warning that occurs when `Enum#native_type` is called.

* Ensure warnings are emitted during spec runs: It appears the use of rspec's `--color` option also disables warnings. It seems this problem doesn't occur with rspec 3.
* Make sure Enum#native_type is tested before modifying it.
* Fix the warning in Enum#native_type and in two places in FFI's specs.

I also noticed both `spec` and `specs` task exist. It *seems* the latter is old and could be removed, but I was unsure so I modified both tasks to emit warnings.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -15,6 +15,7 @@ require 'rubygems/package_task'
 
 RSpec::Core::RakeTask.new(:spec => :compile) do |config|
   config.rspec_opts = YAML.load_file 'spec/spec.opts'
+  config.ruby_opts = '-w'
 end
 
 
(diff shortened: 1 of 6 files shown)


tduehr:
fixed by 18fefbbbd25537e5f458ac794f297a09a0a619f3 and f2e25402359dc6a8f6fa0039bdae579fb0ac9910

mvz:
Great, thanks!

PR #449, opened 2015-08-04
Configure AppVeyor CI

MikaelSmith:
Adds a build script for AppVeyor, issue #444.

PR #450, opened 2015-08-05
Add appveyor.yml for ci-tests on Windows.

larskanis:
This is an extended Version of #449 . It is not ready to merge, because not all ruby versions pass the tests. I'll debug the issues and update this PR accordingly.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -177,13 +177,14 @@ if USE_RAKE_COMPILER
 
   ENV['RUBY_CC_VERSION'] ||= '1.8.7:1.9.3:2.0.0:2.1.5:2.2.1'
 
+  # To reduce the gem file size strip mingw32 dlls before packaging
   ENV['RUBY_CC_VERSION'].to_s.split(':').each do |ruby_version|
-    task "copy:ffi_c:i386-mingw32:#{ruby_version}" do |t|
-      sh "i686-w64-mingw32-strip -S #{BUILD_DIR}/i386-mingw32/stage/lib/#{ruby_version[/^\d+\.\d+/]}/ffi_c.so"
+    task "build/i386-mingw32/stage/lib/#{ruby_version[/^\d+\.\d+/]}/ffi_c.so" do |t|
+      sh "i686-w64-mingw32-strip -S build/i386-mingw32/stage/lib/#{ruby_version[/^\d+\.\d+/]}/ffi_c.so"
     end
 
-    task "copy:ffi_c:x64-mingw32:#{ruby_version}" do |t|
-      sh "x86_64-w64-mingw32-strip -S #{BUILD_DIR}/x64-mingw32/stage/lib/#{ruby_version[/^\d+\.\d+/]}/ffi_c.so"
+    task "build/x64-mingw32/stage/lib/#{ruby_version[/^\d+\.\d+/]}/ffi_c.so" do |t|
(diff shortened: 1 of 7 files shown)


Rakefile
```diff
@@ -177,13 +177,14 @@ if USE_RAKE_COMPILER
 
   ENV['RUBY_CC_VERSION'] ||= '1.8.7:1.9.3:2.0.0:2.1.5:2.2.1'
 
+  # To reduce the gem file size strip mingw32 dlls before packaging
   ENV['RUBY_CC_VERSION'].to_s.split(':').each do |ruby_version|
-    task "copy:ffi_c:i386-mingw32:#{ruby_version}" do |t|
```
tduehr: I think these rake tasks are for cross compiling the gems on linux for release.

spec/ffi/function_spec.rb
```diff
@@ -64,14 +64,14 @@ class << self; self; end
 
   it 'can wrap a blocking function' do
```
tduehr: This test is broken; timing for a blocking call cannot be relied upon within any sane window. see #397

tduehr:
I don't think the two tests you added are being called.

MikaelSmith:
Do you have an AppVeyor environment to test in?

Rakefile
larskanis: Yes they are. But unfortunately they are triggered for native builds, too. So it breaks, because `x86_64-w64-mingw32-strip` is typically not available on Windows. I have verified that, with this patch, the stripping still works per `rake gem:windows` and that it is not called for `rake compile` on Windows.

spec/ffi/function_spec.rb
larskanis: It is less broken than the current. The current implementation does test nothing. It succeeds with :blocking=>true and equally with :blocking=>false.

However this commit is not related to the appveyor result, and should be in a separate PR. I'll remove it here. Anyways it's probably better to test for concurrency, than for any timing...

larskanis:
> I don't think the two tests you added are being called.

Which tests are not called?

I'm curious about the duplicated files in libtest and spec/ffi/fixtures directories. Why are they stored twice?

@MikaelSmith Test results are [here](https://ci.appveyor.com/project/larskanis/ffi).

Rakefile
tduehr: I see it now. Should this be a file task then?

larskanis:
> I don't think the two tests you added are being called.

You're probably speaking about the two tests in spec/ffi/fixtures, don't you? I added these in [this](https://github.com/ffi/ffi/commit/8410d3d1279a72) and [this](https://github.com/ffi/ffi/commit/0499df7ed35cc3) commit, but obviously to the wrong directory. They are specific to windows and are not called on Unix systems.

Rakefile
larskanis: No, it shouldn't. It's an in-place operation, with no proper timestamp to compare.

MikaelSmith:
As a note, AppVeyor could be enabled on ffi/ffi to only run on branches with appveyor.yml, which would enable it on this PR.

tduehr:
You synced the two directories. One is still there only because I keep forgetting to take the time to figure out which is still needed and which can be removed.

tduehr:
I have a personal one I started a while ago. I'll build one for the ffi group later.

Iristyle:
:+1: looking forward to this one getting merged.

ferventcoder:
Also looking forward to this one. :)

tduehr:
Can you pull the blocking call test change? Otherwise, this checks out on my side.

PR #451, opened 2015-08-06
Fix test for :blocking=>true with a sleep()-less version.

larskanis:
This is my attempt to resolve the issue. It is WIP, because it's missing a Windows part.

It tests for concurrency instead of timing. It however has a timeout, so that the test doesn't hang in
a failure condition.

This fixes: https://github.com/ffi/ffi/pull/406

### DIFF
--- a/spec/ffi/fixtures/FunctionTest.c
+++ b/spec/ffi/fixtures/FunctionTest.c
@@ -11,10 +11,10 @@
 #ifndef _WIN32
 #include <unistd.h>
 #include <pthread.h>
+#include <stdlib.h>
 #endif
 
-#include <stdio.h>
-#include <stdlib.h>
+#include "PipeHelper.h"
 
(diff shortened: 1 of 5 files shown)


tduehr:
Does it fail if you set blocking to false?

larskanis:
I now added Windows support for the blocking test as well. Unfortunately it turned out, that there is only little code that I could share between POSIX and Windows. Also the Windows port is more cumbersome, so I separated the two ports into two files plus PipeHelper.h file. This is now quite clean and logical, but takes more lines of code than I thought at first.

larskanis:
@tduehr  Yes, it fails with :blocking=>false on both Windows and Posix. And for :blocking=>false it doesn't hang, but times out after 10 seconds with a failing expectation, on both platforms. And the test takes only 1 milli second on success.

PR #458, opened 2015-09-11
Fixed ino_t, and off_t sizes for x86_32 architectures.

majioa:
Type ino_t has been defaulted to ulong (for most cases), and type off_t defaulted to long type for  x86_32 (32-bit) architectures.


tduehr:
Did you run the rake task to regenerate these files?

majioa:
@tduehr no, I just saw the sources.

PR #468, opened 2015-11-25
minor documentation and code style improvements

paddor:
* fix some YARD syntax errors
* break some monster lines into shorter
* get rid of some empty (but containing blanks) lines


paddor:
@enebo Thanks for pointing this out. Fixed.

PR #469, opened 2015-11-25
FFI::AutoPointer::Releaser#call: call #free instead of #release

paddor:
I'm pretty sure that #free should be called instead of #release.

#free will conditionally call #release (if `@ptr` is set) and, if so, also set `@ptr` and `@proc` to `nil` and `@autorelease` to `false`.


paddor:
Well, okay. If #call only gets called when the associated AutoPointer is getting GC'd, then I guess it's okay to just call #release and save assigning nil/false to those instance variables in the releaser.

PR #479, opened 2016-01-20
Support ruby 2.1.8

ksss:
On ruby 2.1.8, this code is fail. (2.1.7 is success)

```ruby
require 'ffi'

class Foo < FFI::Struct
  layout foo: :bar
end
```

```
.rbenv/versions/2.1.8/lib/ruby/gems/2.1.0/gems/ffi-1.9.10/lib/ffi/struct.rb:344:in `hash_layout': Ruby version not supported (RuntimeError)
```

And current ffi testing code, skip this check since this change https://github.com/ffi/ffi/commit/f75b924ff3abd4c80c66111f1f9c2bbeeb70a171#diff-83f16e6ed1c44a9ced98f69d4139c564

```ruby
"2.1.8" =~ /1\.8\.*/ #=> 2
"2.1.8" =~ /1.8.*/ #=> 2
```

I expect to work `hash_layout` on ruby 2.1.8

### DIFF
--- a/lib/ffi/struct.rb
+++ b/lib/ffi/struct.rb
@@ -341,7 +341,7 @@ module FFI
       # @raise if Ruby 1.8
       # Add hash +spec+ to +builder+.
       def hash_layout(builder, spec)
-        raise "Ruby version not supported" if RUBY_VERSION =~ /1\.8\.*/
+        raise "Ruby version not supported" if RUBY_VERSION =~ /^1\.8\..*/
         spec[0].each do |name, type|
           builder.add name, find_field_type(type), nil
         end
--- a/spec/ffi/struct_spec.rb
+++ b/spec/ffi/struct_spec.rb
@@ -7,7 +7,7 @@ require File.expand_path(File.join(File.dirname(__FILE__), "spec_helper"))
 
 describe "Struct aligns fields correctly" do
   it "char, followed by an int" do
-    pending("not supported in 1.8") if RUBY_VERSION =~ /1.8.*/
+    pending("not supported in 1.8") if RUBY_VERSION =~ /^1\.8\..*/
     class CIStruct < FFI::Struct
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -8,7 +8,7 @@ rvm:
   - "1.9.3"
   - "1.8.7"
   - "2.0.0"
-  - "2.1"
```
tduehr: This change isn't needed. Since travis uses rvm, a version spec of 2.1 will pick up the latest 2.1.x.

lib/ffi/struct.rb
```diff
@@ -341,7 +341,7 @@ def find_type(type, mod = enclosing_module)
       # @raise if Ruby 1.8
       # Add hash +spec+ to +builder+.
       def hash_layout(builder, spec)
-        raise "Ruby version not supported" if RUBY_VERSION =~ /1\.8\.*/
```
tduehr: Please decouple from rubygems. Correcting the regex to `/^1\.8\..*/` is probably a better route.

.travis.yml
ksss: 2 days ago, Travis built by 2.1.5 https://travis-ci.org/ffi/ffi/jobs/103122916

lib/ffi/struct.rb
ksss: OK, squashed.

johnsyweb:
:+1: with thanks.

lib/ffi/struct.rb
johnsyweb:
```ruby
if RUBY_VERSION.start_with?('1.8.')
```
may be clearer.

koic: How is this?

```ruby
Gem::Version.create(RUBY_VERSION) < Gem::Version.create('1.9')
```

PR #480, opened 2016-01-25
Drop Ruby 1.8.7 support

koic:
[It seems that Ruby FFI does not support Ruby 1.8.7 and earlier.](https://github.com/ffi/ffi/pull/479/files)

I propose that required ruby version will be specified to `gemspec` file.

Thanks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ os:
   - osx
 rvm:
   - "1.9.3"
-  - "1.8.7"
   - "2.0.0"
   - "2.1"
   - "2.2.3"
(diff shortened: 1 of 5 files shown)


tduehr:
There's a bit more to it than just updating the gemspec. We'll also want to remove code that handles 1.8.x specifically. Like that which caused #479.

koic:
Thank you for your explanation.
Would you like me to close this issue?

tduehr:
No, leave it open... If you can add to it, do so. Otherwise I'll continue from here at some point...

koic:
I see. I removed code that handles 1.8.x specifically in [this commit](https://github.com/koic/ffi/commit/d18a5fcfb890b5b68d215377c50cd8608ba53d43).
Would you be willing to confirm the codes?

PR #466, opened 2015-11-20
Update to libffi-3.2.1

yghannam:
Signed-off-by: Yazen Ghannam yazen.ghannam@linaro.org

Update to the latest version of libffi. This is needed for AArch64 (and other platforms) support.

rake test results:
Finished in 4.52 seconds (files took 21.48 seconds to load)
4837 examples, 0 failures

PR #483, opened 2016-02-19
Update bundled libffi to 3.2.1

ruscur:
The bundled libffi does not support newer architectures such as ppc64le.
Updating libffi solves this problem.  There were some build issues with
Ruby 2.1 and 2.2 with clang, these were resolved by installing makeinfo
before building.  This also required manually specifying "bundle install",
since Travis did it automatically when "install" was not specified.


ruscur:
This fixes the Travis CI build issues found in #466

ruscur:
Identified some issues, will respin

PR #484, opened 2016-02-19
Update to libffi-3.2.1

ruscur:
This PR applies a .travis.yml build fix on top of yghannam's libffi update (PR #466).  It allows newer architectures such as AArch64 and ppc64le to use the bundled libffi, and still passes all the required Travis CI tests by fixing dependencies.


tduehr:
Instead of installing texinfo, skip building the documentation if possible.

ruscur:
Yeah, that'd be better.  It doesn't look like libffi has a way to do this itself yet:
https://github.com/atgreen/libffi/issues/111

I'll see if I can find a way to do it without touching libffi itself.

ruscur:
I don't think there's an easy fix here.  From what I've seen, Travis CI has no consistency with whether texinfo is present or not, failures are popping up on different configurations (i.e. sometimes on Ruby 2.0 with GCC, other times with Ruby 2.1 with clang, for the same reason: missing makeinfo).  This could be worked around by touching libffi itself, which makes it hard to update the bundled libffi in future.

Would only installing texinfo if it's missing be a better compromise?

tduehr:
It's not just a question of what travis does but what will be needed to get it to compile as part of the gem in general. It should just be doc.. What all needs to be done to remove the makeinfo dependency?

ruscur:
I'm honestly at a loss as to what's going on here.  Locally, it works successfully without makeinfo installed.  On Travis, however, it just fails sometimes.

Sometimes it works with makeinfo missing i.e.: https://travis-ci.org/ffi/ffi/jobs/110848534
yet, mysteriously: https://travis-ci.org/ruscur/ffi-1/jobs/110848517

For the same code, Ruby 2.1 with gcc...

I've gotta be missing something but I have no clue what it is.

ruscur:
Okay, I'm pretty sure this is working now.

Would I need to update libffi.bsd.mk and libffi.darwin.mk as well?  Is "hardcoding" the --disable-docs in there fine or should I pull it out somewhere?

ruscur:
...never mind, yet again I am tricked by travis getting lucky when it still has the chance to intermittently fail.  The --disable-docs option is only present in recent libffi (i.e. not a 2 year old point release, of which there's none newer).

@tduehr - would you prefer updating the bundled libffi to the latest libffi source, or some hacky way of touching the libffi makefiles in extconf or something to skip the docs?  Is touching the bundled libffi itself off limits?  Also, what else would need to be done to integrate a newer libffi aside from getting this docs issue fixed?

tduehr:
Go ahead and update to the latest stable commit. It's probably best to remove the tree and replace it with a submodule while you're at it.

PR #490, opened 2016-03-17
remove RB_GC_GUARD_PTR references

### LINKED ISSUE #489: RB_GC_GUARD_PTR on non-GCC compilers fails to link
I have no idea if this is an ffi bug or a ruby bug or both...

You use RB_GC_GUARD_PTR in at least a couple of places, e.g.:

https://github.com/ffi/ffi/blob/36c9248c079150111c892cc11a7694b6490841b8/ext/ffi_c/MethodHandle.c#L233
https://github.com/ffi/ffi/blob/eb120500a5045e27f1f3b4129349df0190fc4d3a/ext/ffi_c/Function.c#L838

The ruby sources in in ruby.h have a comment which sort of implies maybe you shouldn't?

```
/* RB_GC_GUARD_PTR() is an intermediate macro, and has no effect by
 * itself.  don't use it directly */
```

The ruby.h file then defines the various macros in this kinda nasty #ifdef mess here:

```
#ifdef __GNUC__
#define RB_GC_GUARD_PTR(ptr) \
    __extension__ ({volatile VALUE *rb_gc_guarded_ptr = (ptr); rb_gc_guarded_ptr;})
#else
#ifdef _MSC_VER
#pragma optimize("", off)
static inline volatile VALUE *rb_gc_guarded_ptr(volatile VALUE *ptr) {return ptr;}
#pragma optimize("", on)
#else
volatile VALUE *rb_gc_guarded_ptr_val(volatile VALUE *ptr, VALUE val);
#define HAVE_RB_GC_GUARDED_PTR_VAL 1
#define RB_GC_GUARD(v) (*rb_gc_guarded_ptr_val(&(v),(v)))
#endif
#define RB_GC_GUARD_PTR(ptr) rb_gc_guarded_ptr(ptr)
#endif
```

I'm compiling on AIX with XLC and not GCC so for me this becomes:

```
volatile VALUE *rb_gc_guarded_ptr_val(volatile VALUE *ptr, VALUE val);
#define HAVE_RB_GC_GUARDED_PTR_VAL 1
#define RB_GC_GUARD(v) (*rb_gc_guarded_ptr_val(&(v),(v)))
#define RB_GC_GUARD_PTR(ptr) rb_gc_guarded_ptr(ptr)
```

Now the cool thing here is that the symbol rb_gc_guarded_ptr is only defined in that __GNUC__ block, so this is guaranteed to blow up on any non-GCC/non-MSC install if anyone ever uses RB_GC_GUARD_PTR directly.

So can you just s/RB_GC_GUARD_PTR/RB_GC_GUARD/g and it'll work correctly, or is there a reason why you're trying to access that macro directly?

lamont-granquist:
it seems like the ruby authors do not intend for this to be an external
API.

closes #489

### DIFF
--- a/ext/ffi_c/Function.c
+++ b/ext/ffi_c/Function.c
@@ -202,7 +202,7 @@ function_free(Function *fn)
 static VALUE
 function_initialize(int argc, VALUE* argv, VALUE self)
 {
-    
+
     VALUE rbReturnType = Qnil, rbParamTypes = Qnil, rbProc = Qnil, rbOptions = Qnil;
     VALUE rbFunctionInfo = Qnil;
     VALUE infoArgv[3];
@@ -229,14 +229,14 @@ function_initialize(int argc, VALUE* argv, VALUE self)
          *      Function.new(:int, [ :int ], addr, { :convention => :stdcall })
          */
     }
-    
+
     infoArgv[0] = rbReturnType;
     infoArgv[1] = rbParamTypes;
     infoArgv[2] = rbOptions;
     rbFunctionInfo = rb_class_new_instance(rbOptions != Qnil ? 3 : 2, infoArgv, rbffi_FunctionTypeClass);
 
     function_init(self, rbFunctionInfo, rbProc);
-    
+
     return self;
 }
 
@@ -272,12 +272,12 @@ rbffi_Function_ForProc(VALUE rbFunctionInfo, VALUE proc)
             return cbref;
         }
     }
-    
+
     cbTable = RTEST(rb_ivar_defined(proc, id_cbtable)) ? rb_ivar_get(proc, id_cbtable) : Qnil;
     if (cbTable != Qnil && (callback = rb_hash_aref(cbTable, rbFunctionInfo)) != Qnil) {
         return callback;
     }
-    
+
     /* No existing function for the proc with that signature, create a new one and cache it */
     callback = rbffi_Function_NewInstance(rbFunctionInfo, proc);
     if (cbref == Qnil) {
@@ -297,7 +297,7 @@ static VALUE
 function_init(VALUE self, VALUE rbFunctionInfo, VALUE rbProc)
 {
     Function* fn = NULL;
-    
+
     Data_Get_Struct(self, Function, fn);
 
     fn->rbFunctionInfo = rbFunctionInfo;
@@ -342,7 +342,7 @@ function_init(VALUE self, VALUE rbFunctionInfo, VALUE rbProc)
         rb_raise(rb_eTypeError, "wrong argument type %s, expected pointer or proc",
                 rb_obj_classname(rbProc));
     }
-    
+
     fn->rbProc = rbProc;
 
     return self;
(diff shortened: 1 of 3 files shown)


lamont-granquist:
and to be clear i have no idea if this is right or not, i'm looking for feedback...

larskanis:
I don't think this is the correct solution. `RB_GC_GUARD` protects a single `VALUE` from getting removed by the GC. In both ffi-cases `RB_GC_GUARD_PTR` is used with the intention to protect a whole array of `VALUE`s. So using `RB_GC_GUARD` for these memory pointers will protect nothing.

For the ffi method callback path in [Function.c](https://github.com/ffi/ffi/blob/eb120500a5045e27f1f3b4129349df0190fc4d3a/ext/ffi_c/Function.c#L837) it's best to make use of `rb_apply` and store and pass the parameter `VALUE`s in a ruby Array per `rb_ary_new` instead of a plain memory buffer.

For the ffi method call path in [MethodHandle.c](https://github.com/ffi/ffi/blob/36c9248c079150111c892cc11a7694b6490841b8/ext/ffi_c/MethodHandle.c#L233) I think the additional protection of `argv` isn't necessary at all. It is already protected by the calling function. @tduehr What do you think?

lamont-granquist:
@larskanis thanks, I'll look into this today.

BTW,  my editor has automatic whitespace-rage-fixing enabled, here's the diff without the whitespace fixes:

https://github.com/ffi/ffi/pull/490/files?w=0

lamont-granquist:
@larskanis I think this implements what you mentioned.

I'm a little concerned that I don't know if I need to clean up that array or not.

I also don't know if that RB_GC_GUARD call is still needed or not.

lamont-granquist:
I ran this branch through our internal CI today and it passed on everything (AIX, Solaris, FreeBSD, various Linuxes) except for unrelated-looking failures on Windows that I think is due to installing directly from git and not building the windows gem...  Definitely fixes the AIX/xlc problem I was seeing.

larskanis:
@lamont-granquist Looks good to me. Thank you!

There is no need to cleanup `rbParams` - its free'd by the GC. However the former way to ALLOC a memory array required a explicit `free`, which was missing. So you also fixed a memory leak, by the way. Could you please add this fact to the commit message?

The windows build doesn't pass the CI, currently. That's sad but known.

@tduehr I think we (you) should merge this. Agree?

tduehr:
I haven't had a chance to look at it just yet. This is an area of the ruby api I haven't spent much time with so it'll take a bit for me to eval this. So far, it sounds good.

lamont-granquist:
@larskanis doesn't ALLOCA_N allocate on the stack and it gets automatically free'd (going by the similarity of the name to the alloca() call)?

larskanis:
@lamont-granquist You're totally right: it's `ALLOCA_N` not `ALLOC_N` - sorry for leading you into the wrong direction! And in this case `ALLOCA_N` is the best way to allocate the parameters for the callback invocation. That implicitly protects the VALUEs from getting GC'ed and there's no need to free anything. And there's no need to do any protection to the memory pointer. We can safely remove both `RB_GC_GUARD_PTR` and leave everything else as is.

lamont-granquist:
oh, lol...

'k probably won't get to it till monday...

lamont-granquist:
@larskanis @tduehr PR updated, haven't tested yet, but will be kicking off builds...

tduehr:
@larskanis does this look good now?

larskanis:
@tduehr Yes, LGTM now. And sorry again for the confusion I triggered!

PR #493, opened 2016-04-06
add support for ruby 2.3.0 (2.2.0 was released Dec. 25th 2014,…

bestori:
… which is almost a year and a half before this commit).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ rvm:
   - "2.0.0"
   - "2.1"
   - "2.2"
+  - "2.3"
   - "ruby-head"
   - "rbx"
   - "system"
@@ -21,6 +22,7 @@ matrix:
     - rvm: system
     - os: osx
       rvm: "2.2"
+      rvm: "2.3"
     - os: osx
       rvm: ruby-head
     - rvm: "rbx"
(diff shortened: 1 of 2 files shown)


larskanis:
rake-compiler-dock must be updated to version 0.5.x to support ruby-2.3 cross builds.

tduehr:
I'm not sure travis support 2.3 yet either.

larskanis:
@tduehr Yes, travis needs "2.3.0" as version string.

larskanis:
I'll prepare an updated PR addressing the mentioned issues.

tduehr:
bump 2.2 and 2.1 as well, please.

larskanis:
@tduehr What do you mean with bumping versions 2.2 and 2.1?

In case you're talking about `RUBY_CC_VERSION`, the versions can not easily changed, because they directly refer to the cross ruby versions of rake-compiler-dock. And I'll not change these versions because this would break other projects. Luckily there's no need to change the cross ruby versions, because [ABI compatiblity will not change within the same MINOR version](https://www.ruby-lang.org/de/news/2013/12/21/ruby-version-policy-changes-with-2-1-0/) . I'd update cross ruby versions only in case of such an ABI break. (Ideally rake-compiler would accept `RUBY_CC_VERSION=2.1:2.2:2.3`, but that's not the case for now.)

tduehr:
I would expect rake-compiler-dock to have updated to the latest ruby installer versions. But, you're right, they'll still work for the current ones when compiled against the version in there.

larskanis:
I've opened an updated PR in #495 , that passes all travis-ci versions. It also builds binary windows gems for RubyInstaller-2.3.

bestori:
👍

PR #494, opened 2016-04-09
chore(travis): test against Ruby 2.1.10 and 2.2.4. I cannot build ffi on 2.1.10

sanemat:
In travis-ci, 2.1 means 2.1.5, and 2.2 means 2.2.0.
But latest is 2.1.10 and 2.2.4.

I cannot build `byebug` with Ruby 2.1.10.
chore(travis): test against ruby 2.1.10
https://github.com/deivid-rodriguez/byebug/pull/236
https://travis-ci.org/deivid-rodriguez/byebug/jobs/121867657
~~I think this is `ffi`'s issue.~~
This is `rvm`'s or `travis-rubies`'s issue.

ffi v1.9.10

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,8 @@ rvm:
   - "1.9.3"
   - "1.8.7"
   - "2.0.0"
-  - "2.1"
-  - "2.2"
+  - "2.1.10"
+  - "2.2.4"
   - "ruby-head"
   - "rbx"
   - "system"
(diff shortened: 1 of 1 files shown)


larskanis:
It seems this error is due to the missing dependency to libgmp. This is a rvm issue, that can be solved by installing libgmp-dev package. See also here: https://github.com/rvm/rvm/issues/3509

ruby-2.1.10 and ruby-2.2.4 are not yet available for OSX on travis-ci, so that we need to stay at the older versions for now. See PR #495 .

sanemat:
Thanks, this is `rvm`'s or `travis-rubies`'s issue.

PR #501, opened 2016-05-02
Sets the srcdir to ensure build targets are relative

patrobinson:
Without this change we end up linking to an absolute path:

```
config.status: linking
/home/vagrant/ffi/ext/ffi_c/libffi/src/x86/ffitarget.h to
include/ffitarget.h
```

This makes it difficult to package bundled artefacts for distribution and installation to different directories.


patrobinson:
This breaks a lot of things and I don't have time to look at it.

Issue #488, opened 2016-03-05
blocking: true does not work with varargs


andrewdotn:
I think I’ve found a bug where the `blocking: true` option, which releases the GIL when calling an imported function, does not work with varargs functions.

Here is an example that reproduces the problem. Running 100 threads that sleep for 5ms only takes 20ms total with a non-varargs sleep function. With a varargs sleep function, it takes over 600ms.

```ruby
require 'ffi'
require 'ffi/version'
require 'time'

puts "Ruby #{RUBY_DESCRIPTION}"
puts "FFI #{FFI::VERSION}"

IO.write('sample.c', <<EOF)
#include <stdarg.h>
#include <time.h>

void noop() {
}

void sleep_millis(int millis) {
    struct timespec ts;
    ts.tv_sec = 0;
    ts.tv_nsec = millis * 1000 * 1000;
    nanosleep(&ts, NULL);
}

/* A sample varargs function: add up arguments, indicate end of argument
 * list with 0, sleep for sum, and return sum */
int sleep_sum(int n, ...) {
    va_list args;
    int sum = n;
    int arg;

    va_start(args, n);
    while ((arg = va_arg(args, int)) != 0)
        sum += arg;
    sleep_millis(sum);
    return sum;
}
EOF

puts `gcc -W -Wall -fPIC -c sample.c -o sample.o`
puts `gcc -shared -fPIC sample.o -o sample.dylib`

class Sample
  extend FFI::Library
  ffi_lib './sample.dylib'

  attach_function :noop, [], :void, blocking: true
  attach_function :sleep_millis, [:int], :void, blocking: true
  attach_function :sleep_sum, [:int, :varargs], :int, blocking: true
end

def time_it
  start = Time.now.to_f
  yield
ensure
  printf "%.3fs elapased\n", Time.now.to_f - start
end

Sample.noop

puts "Running noop in 100 threads:"
time_it do
  1.upto(100).map do
    Thread.new { Sample.noop }
  end.map(&:join)
end

puts "Running sleep(5ms) in 100 threads:"
time_it do
  1.upto(100).map do
    Thread.new { Sample.sleep_millis(5) }
  end.map(&:join)
end

puts "Running sleep_sum(1ms, 4ms) in 100 threads:"
time_it do
  1.upto(100).map do
    Thread.new { Sample.sleep_sum(1, :int, 4, :int, 0) }
  end.map(&:join)
end

```

On my machine, running OS X 10.11.3, this outputs:
```
[...]

PR #504, opened 2016-06-04
Support blocking: true for varargs calls

andrewdotn:
As reported in #488, `blocking: true` does not work for varargs calls.

I’ve gotten it working by:
  - Adding a unit test based on the “can wrap a blocking function” spec in `function_spec.rb`
  - Making a typedef and a few functions in `Call.c` public
  - Calling those functions in `Variadic.c` if `HAVE_RB_THREAD_CALL_WITHOUT_GVL` is defined.

This makes `blocking: true` work, and the unit test pass, on modern Rubies. Rubies that require more work to support `blocking: true` will get the old behaviour.

### DIFF
--- a/ext/ffi_c/Call.c
+++ b/ext/ffi_c/Call.c
@@ -338,40 +338,27 @@ rbffi_SetupCallParams(int argc, VALUE* argv, int paramCount, Type** paramTypes,
     }
 }
 
-
-typedef struct BlockingCall_ {
-    rbffi_frame_t* frame;
-    void* function;
-    FunctionType* info;
-    void **ffiValues;
-    void* retval;
-    void* params;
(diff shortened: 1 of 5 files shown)


andrewdotn:
Thoughts?

larskanis:
This all looks good to me! 👍

tduehr:
HAVE_RB_THREAD_CALL_WITHOUT_GVL is to support ruby < 2.0 right?

If so, you can kill off that part of the code.

Can you take a look at #406 as well? Maybe you can fix up that test too.

larskanis:
I already solved #406 - it can be closed, now, but I don't have the permission to do this.

Isn't it better to remove all the support for ruby <2.0 in a separate PR, instead of starting at this by accident?

andrewdotn:
Thanks everyone! It’s been almost a year since the last gem release, are you planning another soon?

PR #507, opened 2016-07-06
Fix library name for symbols

G0dwin:
`libname` needs to be converted to a string whenever it is concatenated, currently I'm getting:

    C:/Ruby22-x64/lib/ruby/gems/2.2.0/gems/ffi-1.9.12-x64-mingw32/lib/ffi/library.rb:129:in `+': no implicit conversion of Symbol into String (TypeError)

This fixes that error.


tduehr:
pulled #508 to fix this

PR #508, opened 2016-07-06
fix library loading by symbol with test

mwrock:
This seems to fix scenarios where the library is provided via a symbol and adds a test for this condition.


mwrock:
Thanks for merging @tduehr ! You think we will see a release soon? This fixes #510

PR #509, opened 2016-07-06
Fix for symbol problems in ffi_lib on windows

jeremydyoung:
I tested the change to ffi/library.rb locally by simply modifying my 1.9.11 gem install.

Unfortunately I don't have a full local C dev environment, so I can't run the rake build to test this out locally.

I think we should have a unit test over this problem, however I don't know if the test I added is worth anything at all.  If I could build it locally, I would definitely run it through its paces, and even make sure I had a failing test or two before submitting this pull request.  Sorry for not having enough ruby / native experience to help out in that area.

Hopefully the travis-ci build will help determine whether I broke something for other platforms.  I'm guessing my new unit test won't actually run there.


jeremydyoung:
I like mwrock's pull request better:  https://github.com/ffi/ffi/pull/508

PR #472, opened 2015-12-03
when use `FFI.typedef :ORITYPE, :TYPEXXX` define a type alisname

windwiny:
when use `FFI.typedef :ORITYPE, :TYPEXXX` define a type alisname

    FFI::AbstractMemory instance can respond_to? [:put_TYPEXXX, :get_TYPEXXX,
                   :write_TYPEXXX, :read_TYPEXXX,
                   :put_array_of_TYPEXXX, :get_array_of_TYPEXXX,
                   :write_array_of_TYPEXXX, :read_array_of_TYPEXXX]

PR #515, opened 2016-07-14
Add AbstractMemory#get and AbstractMemory#put

furunkel:
Similar to #472, but a cleaner and more general approach, IMHO.
These methods are especially useful for writing and reading to pointers of typedef'd types, such
as `size_t`.


tduehr:
Looks good. Can you add some more tests for the other primitive types and for second level typedefs?

furunkel:
Added some more tests. Should now be good to go.

PR #516, opened 2016-07-22
Add Pointer#to_ptr

furunkel:
This adds `Pointer#to_ptr`. It's there in JRuby, and particularly useful to assign struct pointers that 
can possibly be `NULL` to `:pointer` fields, i.e.:

```ruby
  struct_or_null = ...
  other_struct[:pointer_field] = struct_or_null.to_ptr
```

It turns out that the `ffi` gem let's you assign structs directly, without calling `to_ptr`. However, the
call is necessary in JRuby, BTW.

### DIFF
--- a/lib/ffi/pointer.rb
+++ b/lib/ffi/pointer.rb
@@ -130,5 +130,10 @@ module FFI
       }
       self
     end
+
+    # @return [self]
+    def to_ptr
+      self
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


tduehr:
Do you mean:

~~~ruby
other_struct[:pointer_field] = struct_or_null
~~~

If that doesn't work in JRuby, there's a bug there.

furunkel:
@tduehr, yeah. I'll try to come up with a fix for that, along with the new AbstractMemory methods.
Thanks, BTW.

Issue #229, opened 2012-10-03
undefined symbol: ruby_thread_has_gvl_p


rounders:
I'm using ffi-1.0.11 and I'm seeing the following error on heroku (cedar stack, ruby 1.9.3-p194:

unicorn worker[1] -p 54271 -c ./config/unicorn.rb: symbol lookup error: /app/vendor/bundle/ruby/1.9.1/gems/ffi-1.0.11/lib/ffi_c.so: undefined symbol: ruby_thread_has_gvl_p

It just started within the last few days so something likely changed on heroku's platform but I'm at a bit of a loss as to how to track it down. 

I unfortunately cannot update to ffi 1.1.5 as some of the dependencies on this project are locked to ~1.0.7

PR #532, opened 2016-10-18
Create default call frame.

### LINKED ISSUE #527: Problem when embedding Ruby
Hello, I'm embedding ruby in a native program. In this case I'm calling (via rb_require()) a ruby script that creates an object. That object in turn implements a callback (as in https://github.com/ffi/ffi/wiki/Callbacks). I'm doing this via gir_ffi (https://github.com/mvz/gir_ffi).

So, the problem is: since I'm not calling the callback through a Ruby stack frame, rbffi thinks that I don't have the GIL (rbffi_frame_current() -> has_gvl is false). The callback invocation waits for some other thread to release it, which never happns. In short, I'm deadlocked inside `callback_invoke()`.

As a hack, I can call `rbffi_frame_push()` through `dlsym()` (there is no deadlock if I do this) but I'm looking for a proper solution. What can you suggest?

kugel-:
Programs that embed Ruby and use FFI couldn't use FFI callbacks and
call them from the main program because callback_invoke() thought that
the calling thread doesn't have the gvl (even for single threaded programs),
because no call frame was found. There is none because the main program
calls the ruby directly without any ruby function inbetween, This resulted
in a deadlock.

Fixes #527

### DIFF
--- a/ext/ffi_c/Thread.c
+++ b/ext/ffi_c/Thread.c
@@ -344,9 +344,13 @@ thread_data_free(void *ptr)
 void
 rbffi_Thread_Init(VALUE moduleFFI)
 {
+    static rbffi_frame_t root_frame;
 #ifdef _WIN32
     frame_thread_key = TlsAlloc();
 #else
     pthread_key_create(&thread_data_key, thread_data_free);    
 #endif
+    /* Create a root frame for the thread initialising us.
+     * This makes programs that embed ruby work (see #527) */
+    rbffi_frame_push(&root_frame);
 }


tduehr:
I don't think this is the correct fix. I haven't looked at the code that handles this yet but the callback page for the docs makes this look more intricate than this.

kugel-:
Please guide me to the correct fix. From what I can tell it does solve my problem.

Can you follow my analysis of the problem?

tduehr:
Your analysis of the problem and its solution make sense. However, you're looking to solve a very specific problem. I just want to make sure there isn't a better fix that covers a more general case.

larskanis:
I think we need a test case for this specific use case, as well. It should
be possible by building a separate executable binary which makes the
callback into ruby code.

kugel-:
https://gist.github.com/kugel-/20412b414ea5a10ea55079d9b43c82e8 is a pretty self-contained snippet that reproduces the problem. It should be possible to transform it into a test case.

kugel-:
Sorry, https://gist.github.com/kugel-/364fa0011f18f83583f5400f49f6f386 is the correct gist.

larskanis:
@kugel- OK, I can reproduce the issue.

I first wondered if you should include a call to `ruby_init_stack()` or `RUBY_INIT_STACK`, when embedding ruby. But looking into the functions, makes it clear, that these are two independent issues. It has nothing to do with #527 .

Your proposed solution initializes the frame in `rbffi_Thread_Init()`. Therefore the root `rbffi_frame` is for the thread only, that makes the call to `require "ffi"`. Other threads still don't have a `rbffi_frame`. So the assumption is, that it is always this thread, that initially holds the GVL. This sounds reasonable to me, but I'm not entirely sure, that this is always true.

larskanis:
... and the second assumption is, that a callback is called either:

1. from C code within the thread which called `require "ffi"` (the main thread) using the new proposed root `rbffi_frame`
2. from ruby code within the main thread or another ruby thread using the surrounding `rbffi_frame`
3. from a non-ruby thread where the callback will be deferred to a ruby thread

It is certainly also possible to create a ruby thread and run a callback from C without a surrounding `rbffi_frame` . In this case it would still be handled as deferred callback, instead of calling directly into ruby code. However that use case doesn't make much sense to me. And even if it's used, handling as a deferred callback shouldn't deadlock, because there should be another ruby thread, which can receive the callback.

On the other hand, I don't think that the root `rbffi_frame` affects the use case with a standalone ruby version.

To sum up, we should add a MRI-specific test case for embedded ruby (like [this](https://gist.github.com/kugel-/364fa0011f18f83583f5400f49f6f386) ) and merge the proposed change.

kugel-:
Nice to hear, thank you!

I'd like to add that the root `rbffi_frame` shouldn't change anything for existing ffi-using code, because that always pushes one `rbffi_frame` on top.

I'd also like to add to the analysis: My problematic use case is single threaded and the main thread does have the GVL, but ffi doesn't know about it. ffi creates the async_cb_event thread (I see this in gdb) and waits for it to handle the cb. However, it doesn't release the GVL before waiting (again, since ffi thinks it doesn't have the GVL), and so the async cb thread blocks forever while trying to aquire the the GVL. So both threads are waiting for each other, which is the deadlock I'm experiencing.

One way to to fix is letting ffi know the main thread has the GVL, which is achieved with my patch.

Another potential fix could be to unconditionally release the GVL (put pthread_cond_wait() in callback_invoke() inside rb_thread_call_without_gvl()) but that would probably not work with non-ruby thread (Ruby aborts if a thread which doesn't own the GVL attempts to release it doesn't it)?

tduehr:
I would expect this case to normally fall into 3 above... Does this work if the ruby vm is running in a separate thread?

Also, would it be better to push/pop a frame around the callback invocation itself?

larskanis:
> I would expect this case to normally fall into 3 above

Yes, this is the current situation. However this doesn't work, because the main thread holds the GVL, but the ruby VM isn't running continuous. Especially while the C-callback [is called](https://gist.github.com/kugel-/364fa0011f18f83583f5400f49f6f386#file-embed-c-L60) there is no ruby VM running, which could release the GVL, so that the callback-thread could process the deferred callback. Therefore it deadlocks.

With the proposed patch we introduce the use case 1 above, so that the ruby VM is called directly - not by a deferred invocation, which would deadlock.

> Also, would it be better to push/pop a frame around the callback invocation itself?

The `rbffi_frame` is used to decide, how the callback can be delivered to the ruby VM. I don't know how we should decide this without a previously initialized frame struct.

larskanis:
> I would expect this case to normally fall into 3 above

Yes, this is the current situation. However this doesn't work, because the main thread holds the GVL, but the ruby VM isn't running continuous. Especially while the C-callback [is called](https://gist.github.com/kugel-/364fa0011f18f83583f5400f49f6f386#file-embed-c-L60) there is no ruby VM running, which could release the GVL, so that the callback-thread could process the deferred callback. Therefore it deadlocks.

With the proposed patch we introduce the use case 1 above, so that the ruby VM is called directly - not by a deferred invocation, which would deadlock.

> Also, would it be better to push/pop a frame around the callback invocation itself?

The `rbffi_frame` is used to decide, how the callback can be delivered to the ruby VM. I don't know how we should decide this without a previously initialized frame struct.

kugel-:
I experimented a bit with threads. Here's an modified version of the above snippet which uses a separate thread to call the callback: https://gist.github.com/kugel-/7918c1e87f38c0a09d303b93a71927dd

This works with this patch. As expected separate thread uses the "defer callback to a ruby thread" path and works fine.

What does *not* work, though, is when I do the same with a ruby thread `rb_thread_create()`. The problem is fundamentally the same that my patch fixes for the main thread: When the newly created thread runs it owns the gvl (obviously, since it's a ruby thread). `callback_invoke()` again doesn't know about it and uses the "defer callback to a ruby thread" path which then deadlocks upon aquiring the GVL. I don't know how to fix that one. I don't suppose it's possible hook into Ruby's thread creation if/when rb_thread_create() is called from native code? Does pthread offer some hooks, perhaps?

kugel-:
For additional threads there is a workaround available: One can setup the thread function make a detour into Ruby, and then back into C to the actual thread start function (e.g. `rb_thread_create`() -> (C) trampoline function -- (via `rb_funcall`) --> (Ruby) trampoline function -- (via `FFI::attach_function`) --> (C) `rb_rescue` -> (C) actual, user defined thread function (which in my case calls into Ruby via callback).

This inserts a true ruby method into the call stack, and thereby initializing the `rbffi_frame` for that thread before any callback can be called. I positively tested this workaround.

So the main thread can be solved with my patch and additional threads either work already (if using pthread_create or similar) or have a workaround (if using `rb_thread_create`), so please merge my patch.

larskanis:
I've pushed another way to solve the GVL and threading issue [here](https://github.com/ffi/ffi/compare/master...larskanis:fix-527-v1?w=1) . This tests for the GVL state independently to how the call into C code was done. This is also the way how [fiddle checks the GVL state](https://github.com/ruby/ruby/blob/trunk/ext/fiddle/closure.c#L192) . I've checked, that it still works on rubinius, although rbx is broken on travis-ci.

kugel-:
Seems to work for all three cases I tested on Ruby 2.3.1 (1) callback via main thread, 2) callback via non-ruby thread (`pthread_create())`), 3) calback via native ruby thread `rb_thread_create()`).

Was worried to use the functions you make use of, because of #229 which suggested that they cannot be dependended upon.

However, the replacement functions you created when they are not available stuff suffer from the same problem, don't they?

kugel-:
What's the resolution on this? @larskanis your approach is fine to me since I only really care about MRI at the moment.

kugel-:
@larskanis @tduehr ping

larskanis:
pong.

I was too busy with work in order to complete this topic. There is one
issue with my proposal above: the check for ruby_thread_has_gvl_p doesn't
work correctly on MacOS:
https://travis-ci.org/larskanis/ffi/builds/169967964 . The build check in
extconf.rb is successful but calling the function fails, later on.

The MacOS issue could be solved by either
- do a more comprehensive test for ruby_thread_has_gvl_p in extconf.rb or
- do a ruby version check in addition to have_func in extconf.rb .

What do you think?

kugel-:
The function isn't found (at runtime). How's this possible if the extconf.rb check succeeds? I mean, how does extconf.rb find the function?

larskanis:
By building an executable which links to the function. It links
successfully, but this executable doesn't really call the function. Usually
successful linkage is enough indication whether the function is callable,
but in this case it isn't, obviously.

kugel-:
I don't have access to a MAC so I'm afraid I can't help with that. However, I need my problem to be solved :/

kugel-:
What can I do to resolve this issue?

larskanis:
You can do something. I'm currently busy with [RubyInstaller2](https://github.com/larskanis/rubyinstaller2) in my less spare time, but if you like, you can resolve this issue.

OMHO [my proposal](https://github.com/ffi/ffi/compare/master...larskanis:fix-527-v1?w=1) can be used as a starting point, but we need tests for it and the MacOS issue must be solved.

The MacOS issue is easy to solve by issuing `have_func('ruby_thread_has_gvl_p')` only if it's not running on MacOS or the Ruby version is >2.3 where `ruby_thread_has_gvl_p` is exported for sure (for the use in fiddle).

The easiest way to test the call frame issue is probably to mix ffi and fiddle. It should work in both directions ffi->C->fiddle-callback and fiddle->C->ffi-callback. This should be a good indication whether it works with embedded ruby as well.

The other way would be to build an executable with embedded Ruby and run tests with it. However this doesn't fit into the test suite so well.

larskanis:
@tduehr This PR is no longer current. @kugel- has added code and test cases to this issue based on some initial work of mine in PR https://github.com/larskanis/ffi/pull/1 . I'll review and tweak this soon and hope we'll get it ready within the next week.

tduehr:
Closing this one then. Thanks for hunting this down for me.

PR #533, opened 2016-10-18
Updated README.md

srse:
added requirements hint for linux systems running with PaX (see https://github.com/djberg96/sys-filesystem/issues/17 for details)

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,8 @@ At a minimum, you will need:
 * A C compiler (e.g. Xcode on OSX, gcc on everything else)
 * libffi development library - this is commonly in the libffi-dev or libffi-devel
 
+On Linux systems running with PaX (Gentoo, Alpine, etc.) you may need to disable mprotect for ruby (`paxctl -m [/path/to/ruby]`).
+
 ## Installation
 
 From rubygems:
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ At a minimum, you will need:
 * A C compiler (e.g. Xcode on OSX, gcc on everything else)
 * libffi development library - this is commonly in the libffi-dev or libffi-devel
 
-On Linux systems running with PaX (Gentoo, Alpine, etc.) you may need to disable mprotect for ruby (`paxctl -m [/path/to/ruby]`).
+On Linux systems running with [PaX](https://en.wikipedia.org/wiki/PaX) (Gentoo, Alpine, etc.) you may need to disable [mprotect](https://en.wikibooks.org/wiki/Grsecurity/Appendix/Grsecurity_and_PaX_Configuration_Options#Restrict_mprotect.28.29) for ruby (`paxctl -m [/path/to/ruby]`).
 
 ## Installation
 
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ At a minimum, you will need:
 * A C compiler (e.g. Xcode on OSX, gcc on everything else)
 * libffi development library - this is commonly in the libffi-dev or libffi-devel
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -43,6 +43,8 @@ At a minimum, you will need:
 * A C compiler (e.g. Xcode on OSX, gcc on everything else)
 * libffi development library - this is commonly in the libffi-dev or libffi-devel
 
+On Linux systems running with PaX (Gentoo, Alpine, etc.) you may need to disable mprotect for ruby (`paxctl -m [/path/to/ruby]`).
```
simi: PaX and mprotect are new terms for me. Can you add some links please?

tduehr:
https://wiki.archlinux.org/index.php/PaX

Can we get a copy of the violations linux-grsec sent to the kernel log?

I would rather work within the bounds of PaX than tell people to disable it.

srse:
@simi I added some links in the readme as well ;-)

srse:
Sure, the entry in /var/log/kern.log is:

srse:
sure the entry in /var/log/kern.log is (machine specifics masked with ***):
Oct 21 18:34:09 *** kernel: *** grsec: From ***: denied RWX mprotect of <anonymous mapping> by ***/ruby-1.9.2-p330/bin/ruby[ruby:31586] uid/euid:***/*** gid/egid:***/***, parent /bin/bash[sh:31578] uid/euid:***/*** gid/egid:***/***

srse:
I doubt that too, but my PaX knowledge is fairly basic still. Other programming languages and libs trigger the same PaX reaction. Maybe somebody from the @gentoo project can help out with some expert knowledge here? (e.g. @asarubbo or @pacho2)

asarubbo:
The commit message we are talking about says:
```
On Linux systems running with PaX (Gentoo, Alpine, etc.) you may need to disable mprotect for ruby
```
For me, the word 'may' is ambiguous.You should mention where it works or where it doesn't

Anyway, a good hardening expects an hardened kernel(grsec,pax) and the hardening from the toolchain.
In the past I discovered bugs where there is an hardened kernel but not an hardened toolchain (in other words the binary was not PIE).
In which environment it was tested?
If there is a real testcase/reproducer I can give it a try.

srse:
I installed ruby-1.9.2 through rvm on a gentoo with a hardened kernel and then tried to run some ruby code that indirectly requires ffi which failed at the require statement `require 'ffi'` because it failed on an attach_function call within ffi, e.g. if requiring scrypt, it will trigger PaX with an mprotect violation: on this find in this class: https://github.com/pbhogan/scrypt/blob/master/lib/scrypt/scrypt_ext.rb
I suppose in that toolchain the c lib that the scrypt ruby gem wraps around is not position independent and therefore triggering PaX? (if so, it would have to be fixed in any ruby lib that dynamically loads c libs?)

asarubbo:
I don't understand much about ruby, but I downloaded the mentioned script and I tried to execute it:
```
# ruby scrypt_ext.rb 
/usr/lib64/ruby/site_ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file -- ffi-compiler/loader (LoadError)                                                  
        from /usr/lib64/ruby/site_ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require'                                                                                               
        from scrypt_ext.rb:2:in `<main>'
```

Is that the expected error? Am I missing something?

srse:
to reproduce the problem with scrypt:

1. install scrypt gem into ruby gemset: `gem install scrypt`
2. start interactive ruby session: `irb`
3. in irb: `require 'scrypt'

Step 3 will generate the following error:
```bash
There was an error while trying to load the gem 'scrypt'.
Gem Load Error is: ��m�
Backtrace for gem load error is:
***/gems/ffi-***/lib/ffi/library.rb:263:in `attach'
***/gems/ffi-***/lib/ffi/library.rb:263:in `attach_function'
***/gems/scrypt-***/lib/scrypt.rb:14:in `<module:Ext>'
***/gems/scrypt-***/lib/scrypt.rb:12:in `<module:SCrypt>'
***/gems/scrypt-***/lib/scrypt.rb:10:in `<top (required)>'
```

srse:
@asarubbo any luck with reproducing the problem?

PR #548, opened 2017-01-09
Use PRIsVALUE shim when not available for Ruby 1.9.3 compatibility.

### LINKED ISSUE #547: Ruby 1.9.3 compatibility
Commit https://github.com/ffi/ffi/commit/b3eb0f3df668448656e7807085570341fa2d0aa5 breaks compatibility with ruby 1.9.3 as it uses the PRIsVALUE macro which only appeared in ruby 2.0.

Best regards,

Brice

simi:
This fixes #547. It is inspired by https://github.com/ruby/ruby/commit/a1a237fb36d4b78548d2aa94f03a6c460e50b4a7 (also used in https://github.com/ruby/ruby/blob/ff3496b0116ed2ed589d000b7bfca3d8288b009c/ext/json/fbuffer/fbuffer.h#L28-L35).

*I'm not sure if defining of `RB_OBJ_CLASSNAME` is needed here.*

Currently, there's ruby `>= 1.8.7` `required_ruby_version` in gemspec. To support 1.8.7 we need to remove `rb_sprintf` usage also.

It will be nice to decide how to proceed before new version will be released:

1. drop < 2.0 support (change `required_ruby_version` to '>= 2.0.0' in gemspec)
2. merge this and support Ruby >= 1.9.3 (change `required_ruby_version` to '>= 1.9.3' in gemspec)
3. merge this + remove usage of `rb_sprintf` and support => 1.8.7

### DIFF
--- a/ext/ffi_c/AbstractMemory.c
+++ b/ext/ffi_c/AbstractMemory.c
@@ -49,6 +49,14 @@
 #include "Function.h"
 #include "LongDouble.h"
 
+#ifdef PRIsVALUE
+# define RB_OBJ_CLASSNAME(obj) rb_obj_class(obj)
+# define RB_OBJ_STRING(obj) (obj)
+#else
+# define PRIsVALUE "s"
+# define RB_OBJ_CLASSNAME(obj) rb_obj_classname(obj)
+# define RB_OBJ_STRING(obj) StringValueCStr(obj)
+#endif
 
 static inline char* memory_address(VALUE self);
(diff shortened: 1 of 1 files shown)


tduehr:
Direct support for 1.8.7 has been dropped. However, I think it still works so I wasn't planning on updating the gemspec yet. Support for <= 2.0 is deprecated but they still work so it won't be updated past `>= 1.9.3` for now.

PR #553, opened 2017-01-13
Avoid undefined behaviour of LoadLibrary() on Windows.

larskanis:
According to [1] the flag LOAD_WITH_ALTERED_SEARCH_PATH should be used with absolute paths only. It has the effect, that the path of the DLL to be loaded is temporary added to the search path, so that dependent DLLs in the same directory can be found and loaded implicit.

However for relative paths the standard LoadLibrary() search order should be used, because the behaviour of LOAD_WITH_ALTERED_SEARCH_PATH isn't defined for relative paths. In practice (on Windows 10) relative paths to Windows system DLLs do work, so that the library is loaded,
but DLLs in other search paths are not found, when this flag is set.

Ruby's fiddle library uses LoadLibrary without flags in both cases.

[1]
https://msdn.microsoft.com/en-us/library/windows/desktop/ms684179(v=vs.85).aspx

### DIFF
--- a/ext/ffi_c/DynamicLibrary.c
+++ b/ext/ffi_c/DynamicLibrary.c
@@ -185,7 +185,8 @@ dl_open(const char* name, int flags)
     if (name == NULL) {
         return GetModuleHandle(NULL);
     } else {
-        return LoadLibraryExA(name, NULL, LOAD_WITH_ALTERED_SEARCH_PATH);
+        DWORD dwFlags = PathIsRelativeA(name) ? 0 : LOAD_WITH_ALTERED_SEARCH_PATH;
+        return LoadLibraryExA(name, NULL, dwFlags);
     }
 }
 


larskanis:
I've opened the same issue on https://github.com/jnr/jffi/pull/45 for JRuby.

larskanis:
@headius Can you merge this PR into the release?

larskanis:
... It's a blocker for many FFI based gems in the new MSYS2 based RubyInstaller environment. See https://github.com/larskanis/rubyinstaller2/issues/4

larskanis:
Thanks @tduehr !

daniel-rikowski:
I couldn't get ffi to compile with this change. It looks like an `#include <shlwapi.h>` is missing. Also the linker cannot resolve `PathIsRelativeA` (Missing `shlwapi` library reference?)

I wonder why the tests didn't catch this. Perhaps it is a problem on my machine...

larskanis:
@daniel-rikowski yes this include is missing. This is addressed in #556 but not yet merged.

daniel-rikowski:
Ah, thank you, I didn't see that.

headius:
Hello again! I just merged and released jnr/jffi#45 and will look into this one now.

headius:
Ok I see this has been merged as well but presumably not released.

headius:
@tduehr @enebo @larskanis I don't know who did the win32 binary release last time...ping me and we'll coordinate on getting a new ffi gem out.

PR #562, opened 2017-02-28
Use `rb_gc_mark_values` for marking function parameters

tenderlove:
`rb_gc_mark_values` is more efficient than `rb_gc_mark_locations`.
`rb_gc_mark_locations` is typically used for marking addresses from the
stack.  This means that it must check to see if the pointer is a heap
pointer before marking it.  Since we already know the length of this
VALUE array, and that each pointer in the value array did indeed come
from the GC, we don't need to check whether or not the pointer is good
(because we already know it is).


tenderlove:
I have been informed that I am wrong.  Thanks @mattewd

tduehr:
Any chance I can get your input on #532 ?

PR #568, opened 2017-03-21
Add some Requirements

kunashir:
Today when I deployed server I have got next error:
/usr/bin/ld: cannot find -lgmp
collect2: error: ld returned 1 exit status
checked program was:
And install libgmp3-dev save me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ You need a sane building environment in order to compile the extension.
 At a minimum, you will need:
 * A C compiler (e.g. Xcode on OSX, gcc on everything else)
 * libffi development library - this is commonly in the libffi-dev or libffi-devel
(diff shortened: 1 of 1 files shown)


larskanis:
Thank you for providing this pull request! However there is no direct relationship between ffi and libgmp. Instead this is/was an issue of your ruby installation. Possibly you're a victim of this bug: https://github.com/rvm/rvm/issues/3509

Insofar I don't think it is suitable to add this comment to the README.

seliverstov-maxim:
👍  I had same trouble too. Thank you kunashir

Maybe create FAQ? I think it will be very helpful!

tduehr:
This should have been provided by with the ruby installation.

PR #584, opened 2017-07-26
Query ruby thread and GVL states instead of relying on our call frame for callbacks

### LINKED ISSUE #527: Problem when embedding Ruby

larskanis:
This fixes #527 and other interoperability issues that can happen, when a callback is called without a valid call frame.

The change is enabled for ruby-2.3+ only, since it requires the functions ruby_native_thread_p() and ruby_thread_has_gvl_p().

The creation of a call frame is not (yet) removed, because it still carries callback expections around.

This also adds tests for interoperability with Fiddle and embedded ruby use case. Both are affected by the change to the FFI call frame.


larskanis:
@kugel- FYI

kugel-:
Awesome, thanks

larskanis:
I fixed the failing test on MacOS.

spec/ffi/callback_spec.rb
```diff
@@ -771,3 +771,96 @@ module LibTestStdcall
...
+
+  it "from fiddle to fiddle" do
+    assert_callback_in_same_thread_called_once do |block|
+      func = LibTestFiddle.bind_function(:cbVrV, Fiddle::TYPE_VOID, [], &block)
+      LibTestFiddle.testClosureVrV(Fiddle::Pointer[func.to_i])
+    end
+  end
+
+  # https://github.com/ffi/ffi/issues/527
+  if RUBY_ENGINE == 'ruby' && RUBY_VERSION.split('.').map(&:to_i).pack("C*") >= [2,3,0].pack("C*")
```
tduehr: I think you can just compare the strings here as strings.

~~~
RUBY_VERSION >= "2.3.0"
~~~

larskanis: Since we have seen ruby-2.1.10 and I know you as a very exact character, I thought it's best to compare exactly...

rbx doesn't support fiddle, so that these tests fail. The embed_test seems to work in rbx after some tweeking. May I adjust the tests for compatibility with rbx?

tduehr: fair enough...

changing the tests for rbx would depend on the nature of the changes.

larskanis: I don't understand your last comment (as a non English speaker). Shall I adjust the tests so that they pass on rbx (although it is no longer tested on travis-ci)?

Anyway I already tested this with JRuby (see https://github.com/ffi/ffi/pull/584/commits/047371bd21cec9780c53f563d6661e3e65b4fd9a) , so that the update of the testsuite could be merged there someday.

tduehr:
yeah. That wasn't super clear even to me on rereading it...

Is RBX failing because of incorrect behavior in FFI when running under RBX?

spec/ffi/embed-test/ext/embed.c
```diff
@@ -0,0 +1,93 @@
...
+  VALUE mod = rb_define_module("EmbedTest");
+  rb_define_module_function(mod, "testfunc", testfunc, 0);
+}
+
+static VALUE testfunc(VALUE self)
+{
+  int state = 0;
+  VALUE ret;
+
+  rb_eval_string_protect(script, &state);
```
larskanis: > Is RBX failing because of incorrect behavior in FFI when running under RBX?

It's failing because RBX doesn't know `rb_eval_string_protect()` but only `rb_eval_string()`. However the `state` is important to verify that the embedded script worked. So the program has to be changed, so that the success state is transferred in some other way (`rb_rescue()` or similar).

kugel-:
What's holding this back?

tduehr:
Can you rebase this so it runs against the new travis config?

larskanis:
I rebased to current master, but travis-ci tests on MacOS are still pending after 7 hours.

larskanis:
Travis-CI has finished and it is mostly green.

larskanis:
Sorry, looked at the wrong pull request :-(

larskanis:
OK, rebased and I fixed the issue on Windows.

PR #591, opened 2017-09-18
Fix multiple problems on FreeBSD

asomers:


tduehr:
Does this apply to netbsd and openbsd as well?

asomers:
I've only tested it on FreeBSD.  But my expectation would be that the Makefile change affects all of the BSDs, excluding OSX.  The compiler choice change should only affect FreeBSD and OSX.

larskanis:
Is this PR still valid? Or are the issues on FreeBSD fixed with ffi-1.10.0?

asomers:
I can't reproduce it now.

asomers:
Closing stale PR.

PR #594, opened 2017-11-01
Make FFI work on heroku cedar-14

glitch003:
This small change adds a search path to the library lookup logic.  It specifically adds the "/lib/x86_64-linux-gnu/" path which is where libraries live on heroku cedar-14.

This change keeps backwards compatibility with the existing search paths that were already present :)


trostli:
nice. i just ran into this issue myself

trostli:
Actually the other place where libraries live on Heroku is `/app/vendor`. Can you add that path too? That is where they are installed when using buildpacks

glitch003:
@trostli sure, done

trostli:
Thanks!

tduehr:
That needs to be created specific to the platform currently being used rather than being hard coded to x86.

danielpclark:
Shouldn't a server environment use values provided by ENV?  This would allow any server or system to provide proper configurations which is easy enough to document.

```ruby
# # Example
# ENV['FFI_LIBS'] = "/app/vendor/;/lib/x86_64-linux-gnu/"
path = ['/usr/lib/','/usr/local/lib/']
path += ENV['FFI_LIBS'].to_s.split(/[;,]/)
# # Path is now
# ['/usr/lib/', '/usr/local/lib/', '/app/vendor/', '/lib/x86_64-linux-gnu/']
path.find do |pth|
  File.exist?(pth + libname)
end
```

trostli:
good suggestion @danielpclark

gsdean:
is this getting merged?

headius:
@gsdean I see @tduehr comment above but no follow-up commits, so I'm not sure if this is ready or not.

larskanis:
There is another PR which modifies the search path logic: #616 . To be honest, I would rather remove these hardcoded paths than adding random new ones.

tduehr:
Please rebase this if it's still an issue.

dmtask:
When will this be finished?

larskanis:
This will not be merged. We'll not add architecture specific paths that are wrong on all other architectures.

FFI is not responsible for fixing misconfigurations in the library setup. FFI respects the library paths that the operating system provides. This is /etc/ld.conf and LD_LIBRARY_PATH for instance on Linux.

To the contrary we'll remove the static paths entirely in a future release or limit then to operating systems that really need them.

PR #600, opened 2018-01-10
Assigned but unused variable -s

utilum:
Ruby 2.5 warns about this.

### DIFF
--- a/lib/ffi/library.rb
+++ b/lib/ffi/library.rb
@@ -344,7 +344,6 @@ module FFI
       raise FFI::NotFoundError.new(cname, ffi_libraries) if address.nil? || address.null?
       if type.is_a?(Class) && type < FFI::Struct
         # If it is a global struct, just attach directly to the pointer
-        s = type.new(address)
         self.module_eval <<-code, __FILE__, __LINE__
           @@ffi_gvar_#{mname} = s
           def self.#{mname}


tduehr:
`s` is used two lines below that.

utilum:
Right, sorry, I should have seen that. 

Closing. And yet, I do see `user/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/ffi-1.9.18/lib/ffi/library.rb:347: warning: assigned but unused variable - s`. It seems like the parser misreads the code. Any ideas where to hunt this down?

larskanis:
> Any ideas where to hunt this down?

Yes, use an underscored variable like `_s` .

utilum:
That would suppress the warning, but would not be correct use of the convention, as the variable actually is used. I mean to say that Ruby appears to misidentify this as a case of `Assigned but unused variable`, and where in Ruby can I look to fix that.

tduehr:
It’s all good. It took me a couple reads to see it myself. I almost merged it myself. Thanks for the PR, it’s nice to see more people using FFI and helping improve it.

I don’t think Ruby could properly identify this. It’d have to parse the incomplete code in the ‘module_eval’ argument string.

To do that properly you have to postpone the usage check until the block was run. Unfortunately, that’d also likely cause false negatives as some blocks may never get run.

utilum:
@tduehr I think quite a few people are using FFI, as it's a dependency of `ActiveSupport` via `rb-inotify`.

PR #601, opened 2018-01-15
Add mips64(eb) support, and mips r6 support

wzssyqa:
mips64 is same with mips64el, and we have 4 mips r6 arch:
mipsisa32r6   is the same with mips;
mipsisa32r6el is the same with mipsel;
mipsisa64r6   is the same with mips64;
mipsisa64r6el is the same with mips64el.


larskanis:
Thank you for adding these new platforms. I can't really verify if it works, so that it would be nice, if you could have an eye on it in the future.

PR #540, opened 2016-12-12
Fix PaX

forgottenswitch:
Drop `ClosurePool` in favor of PaX-compatible libffi functions.

```
Finished in 2.41 seconds (files took 11.4 seconds to load)
4845 examples, 0 failures
```

PR #617, opened 2018-02-21
Fix failures on MacOS

larskanis:
Fix failures on MacOS intruduced by #540.

See the commit messages for detailed descriptions.


larskanis:
@tduehr Thank you for reviewing! IMHO it would be better to use a merge commit instead of squashing the commits the next time. It otherwise breaks the relation between the commit message and the source changes. If the commits belong semantically together, I do the squashing before PR'ing. Thank you!

tduehr:
Makes sense, will do from here out. I normally squash stuff cause I'm a sucker for a linear, concise history.

larskanis:
Ok, good to know. In contrast, having been forced to linear history by CVS and SVN until 10 years ago, I really enjoy the offline, octopus merge and rebase capabilities of git.

PR #633, opened 2018-05-29
Add FFI::LastError.win_error

### LINKED ISSUE #632: FFI::LastError.error is not that useful under cygwin
Under cygwin, there is no way to access GetLastError. Patch incoming.

graywolf:
Under cygwin both errno and GetLastError are needed. This
adds second getter named #win_error:

    Linux
        FFI::LastError.error     == errno
        FFI::LastError.win_error == NoMethodError
    Windows
        FFI::LastError.error     == GetLastError
        FFI::LastError.win_error == GetLastError
    Cygwin
        FFI::LastError.error     == errno
        FFI::LastError.win_error == GetLastError

Fixed #632

### DIFF
--- a/ext/ffi_c/LastError.c
+++ b/ext/ffi_c/LastError.c
@@ -49,8 +49,17 @@
 # define USE_PTHREAD_LOCAL
 #endif
 
+#if defined(_WIN32) || defined(__CYGWIN__)
+typedef uint32_t DWORD;
+DWORD __stdcall GetLastError(void);
+void __stdcall SetLastError(DWORD);
+#endif
+
 typedef struct ThreadData {
     int td_errno;
+#if defined(_WIN32) || defined(__CYGWIN__)
+    DWORD td_win_errno;
+#endif
 } ThreadData;
 
 #if defined(USE_PTHREAD_LOCAL)
@@ -126,6 +135,19 @@ get_last_error(VALUE self)
     return INT2NUM(thread_data_get()->td_errno);
 }
 
+#if defined(_WIN32) || defined(__CYGWIN__)
+/*
+ * call-seq: win_error
+ * @return [Numeric]
+ * Get +GetLastError()+ value. Only Windows or Cygwin.
+ */
+static VALUE
+get_last_win_error(VALUE self)
+{
+    return INT2NUM(thread_data_get()->td_win_errno);
+}
+#endif
+
 
 /*
  * call-seq: error(error)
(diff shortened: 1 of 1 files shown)


larskanis:
Makes sense and LGTM. However I would call the new accessor winapi_error to be more precise.

ext/ffi_c/LastError.c
```diff
@@ -49,8 +49,17 @@
 # define USE_PTHREAD_LOCAL
 #endif
 
+#if defined(_WIN32) || defined(__CYGWIN__)
+typedef uint32_t DWORD;
```
larskanis: Why do you declare these functions? This shouldn't be necessary and breaks the MINGW build: https://ci.appveyor.com/project/larskanis/ffi-aofqa/build/1.0.30

graywolf: yeah, I'm stupid, this should obviously be only on cygwin, fixed

graywolf:
I don't have strong opinion on the name either way, so I've renamed it.

tduehr:
Looks pretty good. I'm not sure about the Linux `win_error` raising but Windows `error` not... Not sure what the Windows expectation is here as I work in *nix almost exclusively.

Between api compat and the Linux subsystem in win10, this will need to make it into JRuby as well.

larskanis:
@tduehr It would have been better to not unify `errno` and `GetLastError` to `error`. It seemed logical to me as well - until I though about Cygwin. But now it's there and to keep backward compatibility, I think it's best solved like @graywolf proposed.

tduehr:
Would it be better to leave `error` alone, and add `last_error` and `errno` to split the two interfaces?

larskanis:
@tduehr Makes sense. But `last_error` is too imprecise. I would call it `winapi_error` or maybe `GetLastError` like the C function.

tduehr:
Yeah... I think there may be no proper solution here... We'll also need to make similar changes to `FFI::Errno`

graywolf:
So... should I change to pull request to add `LastError#GetLastError`? `GetLastError` is not very ruby-like name for method, so `get_last_error`?

IgorJorobus:
Hello guys, any news on this? I've checked that under Windows (MinGW), FFI.errno calls aren't accurate.

larskanis:
To finally solve this issue I merged it as proposed by @graywolf . It is now released as ffi-1.10.0.

I would greatly appreciate if someone can take the time to write some simple tests for the new call `FFI::LastError.win_error` and `FFI::LastError.win_error=`.

graywolf:
@IgorJorobus could you try to take a look and write some tests as @larskanis asks? Sadly I no longer have a system with cygwin readily available.

IgorJorobus:
Hi guys, I hope that this [specs](https://github.com/ffi/ffi/pull/662) cut the cake for you. Honestly, I couldn't run the them, just coded them. The rake tasks ask for several things that are missing in my system, isn't enough to just clone the repo, I've tried to fulfill rake needs but couldn't during the time I set for this task. Let me know if I can do something else for you.

PR #637, opened 2018-06-09
Pointer#write_array_of_type did not work

ytaka:
The following example that is written in comments of Pointer#write_array_of_type did not work.

    ptr.write_array_of_type(TYPE_UINT8, :put_uint8, [1, 2, 3 ,4])

I fixed it and made test for it.
Would you consider merging these commits?


larskanis:
Thank you - merged!

PR #638, opened 2018-06-09
Adding MacPorts,Fink,etc search path to /opt/local/lib

comomac:
Currently any projects relying on package manager MacPorts and Fink unable to load libraries due to search path. By adding search path for `/opt/local/lib` will resolve this issue and many other ruby libraries using FFI.

Error example:
`LoadError: Could not open library 'glib-2.0': dlopen(glib-2.0, 5): image not found.`


tduehr:
Could have sworn those linked to /usr/local/ ... o well... We should probably add the ability to configure those paths at runtime too...

PR #645, opened 2018-07-29
Compile gem for Linux as well as Windows

joshuaspence:
Cross-compile the gem for Linux as well as Windows. This means that users of this gem don't need to compile locally upon `gem install`.


tduehr:
Please rebase.

joshuaspence:
I'm not sure that I understand the drawbacks. Is it just an issue with Bundler?

tduehr:
Between the defaults in Bundler and the fact we can't rely on dependencies in Linux it's best to leave distributing binary gems to the distributions.

Issue #660, opened 2019-01-07
Move more logic into Ruby code to use the gem in TruffleRuby


eregon:
TruffleRuby, like other Ruby implementations, should support the FFI gem or equivalent API (so gems depending on `ffi` just work).
The FFI gem [currently assumes](https://github.com/ffi/ffi/blob/d976d4ac9845a008ac8637457565a350328a4b44/lib/ffi.rb#L1) that Ruby implementations other than MRI and Rubinius have their own equivalent FFI replacement.
This is not ideal, as it means those implementations (e.g., JRuby, TruffleRuby) need to maintain exact API compatibility and do not benefit from development here.

Instead, what I would like to do is to reuse the code of the FFI gem, but swap out the lower level parts.
While TruffleRuby does support C extensions, the Truffle framework already has its own optimized FFI called *Truffle NFI*. Truffle NFI uses libffi internally and it seems loading twice libffi would not be ideal.
A natural boundary for the "lower level parts" would be the C extension, and then reuse the Ruby code.

However, currently the FFI gem has quite a lot of C extension code (not counting `libffi`), and also exposes a fairly large API:
```
$ cd ext && cloc --exclude-dir=libffi . && cd ..
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
C                               25           1385           1948           5618
C/C++ Header                    24            303            616            900
make                             6             48             29            170
Ruby                             1             12              4             56
-------------------------------------------------------------------------------
SUM:                            56           1748           2597           6744
-------------------------------------------------------------------------------

$ cloc lib
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
Ruby                            24            394           1335           1519
-------------------------------------------------------------------------------
SUM:                            24            394           1335           1519
-------------------------------------------------------------------------------
```

But a fair part of the C extension code doesn't seem to need to be written in C.
In fact, I started rewriting parts of the C extension to Ruby, and it seems to work well so far.

So I would like to know if PRs moving C extension logic to Ruby would be welcome.
I think it would be good as it would make the gem more maintainable by having clearer and more expressive code than is possible with the C extension API.
Of course, performance is a concern so translating performance-critical parts needs to be evaluated.

I will start with a simple PR moving a class from C to Ruby.

larskanis:
@tduhr do you have any objections? I'm fine with this direction.

[...]

PR #661, opened 2019-01-07
Move FFI::DataConverter to Ruby

eregon:
* Include DataConverter in StructByReference in Ruby code to simplify ordering.

See #660 for context.


lib/ffi/data_converter.rb
```diff
@@ -0,0 +1,37 @@
+module FFI
```
eregon: Should I add a license header? If so, what copyright line should I use?

eregon:
This is now ready for review.

larskanis:
LGTM

eregon:
Can we merge this?

@tduehr Could you review?

I think this doesn't touch performance critical-code, but I'm happy to run some benchmarks if you tell me which ones would be relevant for this.

The behavior seems similar to [the DataConverter class in JRuby](https://github.com/jruby/jruby/blob/4e8bb2666a7257f0f5986800f96bb88efdd6acbd/core/src/main/java/org/jruby/ext/ffi/DataConverter.java), except that one has the extra `reference_required` and `reference_required?` methods.

tduehr:
There's more to it than just the module itself. I'll take a look at usage today.

I expect this code is used in converting from ruby to C/JAva and back... Any benchmark covering that should be good.

Can you get the benchmark suite working, or at least find the holes? That'd be awesome too...

lib/ffi/data_converter.rb
headius: Follow whatever pattern other .rb files use. We can discuss sorting out licensing, headers, and copyrights in a separate issue if necessary.

headius:
@tduehr Even if this is on the hot path it's probably a lateral move changing this code to Ruby. It may even be faster in some cases since the C code (and the Java code, for that matter) does uncached instance variable lookups.

tduehr:
The licensing is all over the place. We'll want to consolidate that at some point.

eregon:
I added the license header.

lib/ffi/data_converter.rb
eregon: I used the same as in `lib/ffi/errno.rb`.

PR #664, opened 2019-01-15
Add tests on armhf and ppc64le linux, JRuby and Truffleruby on Travis-CI

larskanis:
CI testing is currently done on Windows-x86, Windows-x64, Linux-x86_64 and MacOS-x86_64.
This adds PPC64le and ARMv7 by using docker and qemu.

This also adds JRuby and Truffleruby, so that we notify the differences to them. They currently fail, but the aim is to fix JRuby and Truffleruby support in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,20 @@ matrix:
     - os: osx
       rvm: 2.4.2
   include:
+    - name: powerpc
+      language: generic
+      before_install: |
+        docker run --rm --privileged multiarch/qemu-user-static:register --reset &&
+          docker build --rm -t ffi-powerpc -f spec/env/Dockerfile.powerpc .
+      script: |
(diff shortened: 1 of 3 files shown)


tduehr:
Can you separate the platform tests PR out from this? It can be added without further testing.

Can you add a change in something that'll cause a test to fail if the Ruby files here are loaded in jruby or truffleruby? We need positive confirmation that the correct code is being tested.

larskanis:
I switched one docker image to powerpc to have a test on a big endian arch.

> Can you separate the platform tests PR out from this? It can be added without further testing.

Sure, but the JRuby and Truffleruby cases are in `allowed_failures`, currently. I would like to add them to better track our progress.

> Can you add a change in something that'll cause a test to fail if the Ruby files here are loaded in jruby or truffleruby?

Sorry, I don't understand the question!

PR #667, opened 2019-01-20
CI: Drop "system" ruby

olleolleolle:
This PR reduces the size of the CI matrix by removing `system`.

  - its file location for gems is not even writable, so it fails to install
  - it does not represent any new Ruby version which does not already exist in the matrix


larskanis:
Makes sense. Thanks again Olle!

PR #671, opened 2019-01-25
ci tweaks

tduehr:


larskanis:
Thanks! I don't have a Mac, so that I don't know which tests are useful there...

PR #676, opened 2019-01-31
Fix path to libtest in benchmarks

eregon:
cc @tduehr @larskanis

### DIFF
--- a/bench/bench_helper.rb
+++ b/bench/bench_helper.rb
@@ -3,4 +3,4 @@ require "rubygems"
 require 'benchmark'
 require 'ffi'
 ITER = ENV['ITER'] ? ENV['ITER'].to_i : 100000
-LIBTEST_PATH = "#{Dir.getwd}/build/libtest.#{FFI::Platform::LIBSUFFIX}"
+LIBTEST_PATH = File.expand_path("../../spec/ffi/fixtures/libtest.#{FFI::Platform::LIBSUFFIX}", __FILE__)


tduehr:
This doesn't work. I think there needs to be a change in the makefile too.

eregon:
> This doesn't work. 

It does, but `libtest` needs to be compiled first:
```
$ rake
compiles libtest and runs specs

$ ruby bench/bench_IIIrV.rb             
Benchmark [ :int, :int, :int ], :void performance, 100000x calls
  0.010321   0.000081   0.010402 (  0.010420)
  0.010791   0.000000   0.010791 (  0.010822)
  0.010769   0.000000   0.010769 (  0.010798)
  0.010546   0.000000   0.010546 (  0.010563)
  0.010551   0.000001   0.010552 (  0.010570)
  0.010586   0.000000   0.010586 (  0.010610)
  0.010808   0.000000   0.010808 (  0.010839)
  0.010427   0.000008   0.010435 (  0.010447)
  0.010485   0.000000   0.010485 (  0.010497)
  0.009805   0.000953   0.010758 (  0.010786)
Benchmark ruby method(3 arg)  performance, 100000x calls
  0.003196   0.000016   0.003212 (  0.003223)
  0.003231   0.000017   0.003248 (  0.003254)
  0.003106   0.000008   0.003114 (  0.003117)
  0.003099   0.000007   0.003106 (  0.003111)
  0.003140   0.000000   0.003140 (  0.003145)
  0.003076   0.000000   0.003076 (  0.003079)
  0.003076   0.000000   0.003076 (  0.003079)
  0.003127   0.000000   0.003127 (  0.003131)
  0.003148   0.000000   0.003148 (  0.003158)
  0.003194   0.000000   0.003194 (  0.003206)
```

The logic to compile libtest is in `spec/ffi/spec_helper.rb`.
Maybe it could be extracted to its own file and shared by specs and benchmarks (both using libtest).

But in any case, I think it's worth merging this PR as it does work as shown above, unless there is some platform-specific issue?

>  I think there needs to be a change in the makefile too.

Which Makefile?

larskanis:
@eregon You're absolutely right, the libtest build code should be separated into it's own file and `bench_helper.rb` should trigger it. And the current code is broken, so that it's better to merge this than leaving it broken.

PR #677, opened 2019-02-01
Fix benchmarks and harness, compile libtest automatically

eregon:
Following the discussion in https://github.com/ffi/ffi/pull/676.
Now all benchmarks run fine, and importantly use the FFI built in the repo (instead of the installed gem).

This can run all benchmarks in a short time to verify they don't error out:
```
for f in bench/bench_*.rb; do echo $f; ITER=10 ruby $f; done
```

A good next step would be to convert benchmarks to use `benchmark-ips` since currently the number of iterations is arbitrary and often not so well calibrated. I'd rather let FFI maintainers do that though :smiley: 

cc @larskanis @tduehr

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ language: ruby
 script:
   - bundle exec rake compile || bundle exec rake compile
   - bundle exec rake test
+  - ITER=10 bundle exec rake bench:all
 os:
   - linux
   - osx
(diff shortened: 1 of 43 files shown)


larskanis:
Looks all reasonable! Especially adding the benchmarks to Travis-CI.

eregon:
@larskanis @tduehr Can we merge this?

larskanis:
Thanks again for your great work! 👍

PR #681, opened 2019-02-18
Move FFI::StructByReference to Ruby

eregon:
See #660.

It's much shorter and more readable in Ruby:
commit 2d09ca5 removes 239 lines of C and add only 39 lines of Ruby.

I don't expect any of this to be performance critical.
None of the benchmarks use `StructByReference`.
The C code was anyway basically just calling out to Ruby core methods.
Having it in Ruby also allows inline caching which is not possible in C with MRI.

I splitted StructByReference and StructLayout in their own files to ease code navigation/improve readability and reduce the length of `lib/ffi/struct.rb`.

cc @larskanis @tduehr


larskanis:
I read every line of changed code and it all makes sense. Thanks again for your contribution!

eregon:
Thanks for merging quickly :)

PR #692, opened 2019-04-11
Fix FFI::Pointer#read_string(0) to return a binary String

### LINKED ISSUE #691: FFI::Pointer#read_string(0) returns a UTF-8 and not a binary String
This is inconsistent with all other pointer string methods, which return a binary String.

```ruby
$ pry -rffi
[2] pry(main)> FFI::MemoryPointer.new(:char, 3).read_string(0)                                                                              
=> ""
[3] pry(main)> FFI::MemoryPointer.new(:char, 3).read_string(0).encoding                                                                     
=> #<Encoding:UTF-8>
[4] pry(main)> FFI::MemoryPointer.new(:char, 3).read_string(1).encoding                                                                     
=> #<Encoding:ASCII-8BIT>
```

eregon:
* Like other pointer string methods and usages.
* Fixes #691.
* String#b is available since Ruby 2.0.

### DIFF
--- a/lib/ffi/pointer.rb
+++ b/lib/ffi/pointer.rb
@@ -49,7 +49,7 @@ module FFI
     # equivalent string if +len+ is not +nil+.
     def read_string(len=nil)
       if len
-        return '' if len == 0
+        return ''.b if len == 0
         get_bytes(0, len)
       else
         get_string(0)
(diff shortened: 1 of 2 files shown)


eregon:
cc @larskanis

larskanis:
Good catch, makes sense!

eregon:
@larskanis Thanks for merging :)

PR #694, opened 2019-04-27
Fix wiki links in README

f3ndot:
image

wiki.github.com appears to render an error and the canonical URL for GitHub wikis are not a subdomain but in fact a subdirectory /wiki/


larskanis:
Thank you for fixing this!

PR #695, opened 2019-05-16
Add specs for missing cases

eregon:
Along with my mostly pure-Ruby implementation for TruffleRuby (#660), I discovered a few incompatibilities in TruffleRuby which were not covered by existing specs.

So in this PR, I add specs for previously untested code.
cc @larskanis

### DIFF
--- a/spec/ffi/callback_spec.rb
+++ b/spec/ffi/callback_spec.rb
@@ -500,28 +500,32 @@ describe "Callback with " do
     expect(v).to eq(-1)
   end
 
+  def testCallbackU8rV(value)
+    v1 = 0xdeadbeef
+    LibTest.testCallbackU8rV(value) { |i| v1 = i }
+    expect(v1).to eq(value)
+
+    # Using a FFI::Function (v2) should be consistent with the direct callback (v1)
+    v2 = 0xdeadbeef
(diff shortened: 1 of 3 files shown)


spec/ffi/callback_spec.rb
```diff
@@ -500,27 +507,39 @@ class S8F32S32 < FFI::Struct
...
+
+    # Using a FFI::Function (v2) should be consistent with the direct callback (v1)"
+    v2 = 0xdeadbeef
+    fun = FFI::Function.new(:void, [:uchar]) { |i| v2 = yield(i) }
+    LibTest.testCallbackU8rV(fun, value)
+    expect(v2).to eq(value)
+  end
+
   it ":uchar (0) argument" do
     v = 0xdeadbeef
```
larskanis: You can remove this and the `expect` line below as well, can't you?

eregon: Good catch, indeed.

eregon: This is fixed now (I originally wrote the helper a bit differently, but the latest version seems the simplest).

eregon:
@larskanis Could you merge this? :)

eregon:
Thanks!

PR #698, opened 2019-05-17
Complete gemspec metadata

olleolleolle:
In order not to lose any of the links at release, when RubyGems writes them in, again.


larskanis:
Didn't know there is a google group for ruby-ffi. I just subscribed to it... Thanks!

PR #711, opened 2019-06-30
Fix BigDecimal.new

wonda-tea-coffee:
It is as the subject.

Refer
ruby/bigdecimal@5337373


larskanis:
Thanks for fixing this!

PR #712, opened 2019-07-28
Generator spec

jethrodaniel:
Adds a test for the generator, and also fixes a typo in the generator comments.


jethrodaniel:
Not sure why the AppVeyor builds failed - it seems unrelated, though.

jethrodaniel:
Had to install `zlib1g-dev` on both Dockerfiles to use the example from the generator code.

Not sure if that's the way we should go here, since it caused the allowed failure travis builds to fail, maybe we could use a different example, or get the builds working for those failures?

PR #627, opened 2018-04-26
Switch time_t and suseconds_t types to long on FreeBSD

meanphil:
Currently time_t and suseconds_t are ints, which should be longs. This causes problems with various libraries, but especially anything that uses select()

See here: https://github.com/freebsd/freebsd-ports/commit/ed7f235bd792070adcc5d77598bd55e1874ac3d4
And: https://github.com/typhoeus/ethon/issues/98

Issue #544, opened 2017-01-02
time_t and suseconds_t types are wrong for freebsd


dcws:
There is a longstanding bug with the ethons library such that select() intermittently returns 22 on select on FreeBSD, I have just tracked this down to the type definition in ffi being incorrect for (at least) time_t and suseconds_t  in ffi you have them as ints.. However a printf of sizeof(...) on FreeBSD 11-amd64 indicates they are 8 bytes long (and therefore int64).. I updated the specification to long in the files and was just able to solve the errno22 issue.

I am not sure at what point FreeBSD cut over to 64 bits for those structures (it was a LONG time ago though), how you would version differences in ABI formats within a platform, what is the correct value (long vs int64), what the difference is for the 2 time_t and suseconds_t types, or what other values there are incorrect; but I hope this could be quickly updated.

t-zuehlsdorff:
Just to prevent confusion:
I just committed a patch to the FreeBSD ports-tree which will change the types until the bug is fixed in an release. :)

dcws:
This was fixed in 
https://github.com/ffi/ffi/pull/627  thank you!

PR #721, opened 2019-10-09
Update types for FreeBSD 12

adam12:
The types for time_t and suseconds_t should be long and not int.

Complementary of 4ba53c5

Related to #544 and #627


larskanis:
Could you please update ffi to latest changes on master, run `rake types_conf` and send a pull request for the resulting changes? It should update the types based on system header files for platform `x86_64-freebsd12`.

adam12:
I've opened #722 with the generated types. I had to install and use `gcc`. I'm not sure if there will be any drawbacks from that.

I did notice a bunch of new types have been added to the list of types. Let me know if there is anything else required.

larskanis:
This is fixed in libpq and will be part of PostgreSQL-12.1: https://github.com/postgres/postgres/commit/ba19a6b73c5bd771d8864171ede03503a9ff564e

larskanis:
Thanks @adam12 , I merged #722 . I'm curious why clang doesn't work, but I think gcc headers are OK as well.

adam12:
> Thanks @adam12 , I merged #722 . I'm curious why clang doesn't work, but I think gcc headers are OK as well.

Thanks for merging!

PR #724, opened 2019-10-27
identify dragonflybsd as a platform

ahorek:


larskanis:
LGTM. Did you generate the attached `types.conf` per `rake types_conf`? Which compiler did you use?

ahorek:
Did you generate the attached types.conf per rake types_conf - yes
Which compiler did you use? gcc version 8.3 [DragonFly] Release/2019-02-22

PR #730, opened 2019-11-24
Remove taint support

y-yagi:
Ruby 2.7 deprecates taint and it no longer has an effect.
https://bugs.ruby-lang.org/issues/16131

The APIs related to taint cause deprecation warning message and the lack of taint support should not cause a problem in previous Ruby versions.
So I removed that. But if need to taint support that to old versions of Ruby, I will update PR.


larskanis:
Thank you @y-yagi ! I think we can remove taint support unconditionally.

y-yagi:
Thanks!

PR #735, opened 2020-01-08
Disallow struct layout changes

### LINKED ISSUE #734: Disallow re-layout of existing Struct
I ran into this while attempting to align all Ruby code in JRuby and the FFI gem.

Currently, it is possible to define a struct and change its layout repeatedly:

```ruby
class PairLayout < FFI::Struct
  layout :a, :int, :b, :long_long
end
class PairLayout < FFI::Struct
  layout :a, :int, 0, :b, :long_long, 4 # allowed
end
```

However, this seems like not only an odd feature, but a bad/dangerous one. When the layout has been used for any existing objects, changing it will produce very difficult-to-diagnose bugs (at least) or potentially memory violations if the layout is re-retrieved from the class and used for previous pointers.

At some point, it appears this was disallowed but the error was [commented out](bd5cf88787d4cd53b17a77cec66ae20824c2ef1e) for some reason: https://github.com/ffi/ffi/blob/master/lib/ffi/struct.rb#L207

This behavior caused a failure on JRuby running struct_spec.rb, because two specs there open the same `PairLayout` struct and call layout each time. JRuby's implementation of FFI expects that after the first struct has been allocated the layout will not be changed again, and caches it to avoid repeatedly looking up the instance variable. This caching resulted in the first spec's layout being used for the second, causing it to fail.

I believe we should reinstate the error. I cannot think of any good reasons why you would want to re-layout a struct after you have already started creating instances of it, and many good reasons why doing so would be a bad idea.

For now, I have removed the caching in JRuby, but I'd like to revert that (and the C FFI could potentially start caching the layout then as well): https://github.com/jruby/jruby/pull/5948/commits/21a8a8cc9d36b1d15e82cbec1015a50aea7d9137

larskanis:
This seems like not only an odd feature, but a bad/dangerous one.
When the layout has been used for any existing objects, changing it will produce
very difficult-to-diagnose bugs (at least) or potentially memory violations
if the layout is re-retrieved from the class and used for previous pointers.

It was heavily used in the specs to overwrite one and the same class multiple times to change the layout immediatelly before the test.
However this use case is specific to testing and should not be used anyway.

Fixes #734

### DIFF
--- a/bench/bench_FFFrV.rb
+++ b/bench/bench_FFFrV.rb
@@ -1,24 +1,26 @@
 require_relative 'bench_helper'
 
-module LibTest
-  extend FFI::Library
-  ffi_lib LIBTEST_PATH
-  attach_function :bench_f32f32f32_v, [ :float, :float, :float ], :void
-  def self.rb_bench(a0, a1, a2); nil; end
+module BenchFFFrV
+  module LibTest
+    extend FFI::Library
+    ffi_lib LIBTEST_PATH
+    attach_function :bench_f32f32f32_v, [ :float, :float, :float ], :void
+    def self.rb_bench(a0, a1, a2); nil; end
+  end
+
+
+  puts "Benchmark [ :float, :float, :float ], :void performance, #{ITER}x calls"
+  f = 1.0
+  10.times {
+    puts Benchmark.measure {
+      ITER.times { LibTest.bench_f32f32f32_v(f, f, f) }
+    }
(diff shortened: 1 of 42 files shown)


larskanis:
Lots of indents have changed, best view the diff by https://github.com/ffi/ffi/pull/735/files?w=1

headius:
Warning looks good to me.

PR #745, opened 2020-02-15
fix cast between incompatible function types

ahorek:
I have a warning on Windows
```
ClosureTest.c: In function 'testThreadedClosureVrV':
ClosureTest.c:102:44: warning: cast between incompatible function types from 'void * (*)(void *)' to 'void (*)(void *)' [-Wcast-function-type]
  102 |     HANDLE hThread = (HANDLE) _beginthread((void (*)(void *))threadVrV, 0, &arg);
```
gcc version 9.2.0

there's probably a better way how to fix it.

### DIFF
--- a/spec/ffi/fixtures/ClosureTest.c
+++ b/spec/ffi/fixtures/ClosureTest.c
@@ -78,7 +78,11 @@ struct ThreadVrV {
     int count;
 };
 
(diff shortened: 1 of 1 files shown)


larskanis:
Thank you! I took a similar approach.

Issue #743, opened 2020-02-03
ffi 1.12.2 did not push its -java variant


kares:
could you someone please make sure a build using JRuby is also pushed?
https://github.com/ffi/ffi/commit/ad524cb016fc788861d76e43a5067b8d441a76bc // cc @larskanis 

otherwise this is (potentially) breaking JRuby `bundle` install with the ffi as a dependency.

dfighter1985:
Could this be the same reason why I am getting this?

> 14:14:05  [INFO] Building native extensions.  This could take a while...
> 14:14:06  [WARNING] ERROR:  Error installing /home/jenkins/.m2/repository/rubygems/ffi/1.12.2/ffi-1.12.2.gem:
> 14:14:06  [INFO] 8 gems installed
> 14:14:06  [WARNING]     ERROR: Failed to build gem native extension.
> 14:14:06  [WARNING] 
> 14:14:06  [WARNING]     current directory: /home/jenkins/workspace/ce_5.5_OSZK_branches_sworvd2-kds/dspace/modules/xmlui-mirage2/target/rubygems/gems/ffi-1.12.2/ext/ffi_c
> 14:14:06  [WARNING] /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java -cp :/home/jenkins/workspace/ce_5.5_OSZK_branches_sworvd2-kds/dspace/modules/xmlui-mirage2/target/test-classes:/home/jenkins/workspace/ce_5.5_OSZK_branches_sworvd2-kds/dspace/modules/xmlui-mirage2/target/classes org.jruby.Main -r ./siteconf20200203-365-no79qr.rb extconf.rb
> 14:14:06  [WARNING] Error: Could not find or load main class org.jruby.Main
> 14:14:06  [WARNING] 
> 14:14:06  [WARNING] extconf failed, exit code 1

larskanis:
Sorry, I forgot!

PR #747, opened 2020-02-16
Load JRuby ext under JRuby

headius:
JRuby's FFI is almost aligned with the gem via jruby/jruby#5948. This change will allow us to use ffi.rb unmodified.

### DIFF
--- a/lib/ffi.rb
+++ b/lib/ffi.rb
@@ -8,6 +8,10 @@ if !defined?(RUBY_ENGINE) || RUBY_ENGINE == 'ruby' || RUBY_ENGINE == 'rbx'
 
   require 'ffi/ffi'
 
+elsif RUBY_ENGINE == 'jruby'
+  JRuby::Util.load_ext("org.jruby.ext.ffi.FFIService")
+  require 'ffi/ffi'
+
 elsif defined?(RUBY_ENGINE)
   # Remove the ffi gem dir from the load path, then reload the internal ffi implementation
   $LOAD_PATH.delete(File.dirname(__FILE__))


larskanis:
This is awesome Charles! I'm very glad, that MRI and JRuby FFI are aligned now. I'm sorry I can't help with JRuby, but I don't use Java at all. It takes me too much time to do serious work. Anyway I'd like to work together if there are issues with C or Ruby parts.

Just to clarify: This change allows to sync `lib` and `spec` directories without modifications between both repositories? Using the ffi-platform-ruby-gem on JRuby is still not intended, is it?

There seems to be no code for synchronization in https://github.com/jruby/jruby/pull/5948 so far. Is this something you plan to add?

larskanis:
@headius Can you answer the questions above? I'd like to merge this PR.

headius:
@larskanis Sorry I didn't see this comment!

Yes, JRuby's FFI lib and spec should now be exactly the same as from the gem. There's no syncing code in #5948 because my hope is that we can source the gem contents directly, as a default gem.

I have created https://github.com/jruby/jruby/pull/6150 to track that change. It doesn't work properly at the moment because the "java" version of the gem contains no sources. I suggest either we include the sources there, or release a single gem that just skips the extension build on JRuby.

larskanis:
> I suggest either we include the sources there, or release a single gem that just skips the extension build on JRuby.

@headius I merged this manually. A platform=java gem with only ruby sources is much smaller than the ordinary platform=ruby gem. So I changed the java gem accordingly in #763 .

eregon:
How much smaller is it?

For TruffleRuby we'll want to use the platform=ruby gem at some point to reuse the Ruby files from the gem, so I think longer term it would be simpler to only have a platform=ruby gem (also in terms of complexity of the Rakefile, etc).

headius:
I do not have any preference between using the `ruby` gem or the `java` gem, so I'll leave that up to you.

larskanis:
With #763 both platform gems (ruby and java) are equally usable. `extconf.rb` already [compiles the extension conditionally](https://github.com/ffi/ffi/blob/e9dab3c84da437de37b0068fcc98a7b161cb9060/ext/ffi_c/extconf.rb#L3) and building a platform java gem is just a few lines of simple code in the `Rakefile`. So I'm OK with both of them.

I also wouldn't omit the `platform=java` gem since the java version has been [explicit requested](https://github.com/ffi/ffi/issues/743) several times, although ffi-1.12.2 installs pretty fine on JRuby-9.2 for me (it runs a dummy makefile and falls through to the internal JRuby impl at runtime).

Sizes of the gems are as following:
```sh
$ ll pkg/*.gem
-rw-r--r-- 1 lars lars 898048 Mär 30 21:07 pkg/ffi-1.12.2.gem
-rw-r--r-- 1 lars lars  51712 Mär 30 21:07 pkg/ffi-1.12.2-java.gem
```

PR #751, opened 2020-03-16
Make string return value from callback available

datibbaw:
## Background
When a `callback` is defined, the return value isn't made available to the underlying C code if the return value is a `string` memory pointer; instead, it would return `null` to the caller.

## Disclaimer
Tested with two return values:
- `FFI::MemoryPointer.from_string()`
- Regular `string` value (does not work yet)

## Reference
- [SO Question](https://stackoverflow.com/questions/60689128/ruby-ffi-callback-return-values)

### Ruby code
```ruby
require 'ffi'

class Foo
  extend FFI::Library
  ffi_lib File.expand_path('fun.o')

  callback :incoming_rpc, [:string], :string
  attach_function :do_some_work, [:incoming_rpc, :string], :string

  def initialize
    output = do_some_work("Ruby init...") do |name|
      FFI::MemoryPointer.from_string("hello #{name}")
    end
    puts "Output: #{output.inspect}"
  end
end

Foo.new
```
### C code
```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

typedef const char *callbkfn(char *);

extern char* do_some_work(callbkfn fn, char *name);

char* do_some_work(callbkfn fn, char* userdata) {
  printf("do_some_work param: %s\n", userdata);
  printf("callback output: %s\n", fn(strdup("Hello from C")));
  return strdup("Returned from C");
}
```

### DIFF
--- a/ext/ffi_c/Function.c
+++ b/ext/ffi_c/Function.c
@@ -843,10 +843,11 @@ invoke_callback(VALUE data)
         case NATIVE_STRING:
             if (TYPE(rbReturnValue) == T_DATA) {
                 *((void **) retval) = ((AbstractMemory *) DATA_PTR(rbReturnValue))->address;
-            } else if (TYPE(rbReturnValue) == T_STRING) {
-                VALUE obj = rbffi_MemoryPointer_ForString(rbReturnValue);
-
-                *((void **) retval) = ((AbstractMemory *) DATA_PTR(obj))->address;
+//            This causes the block to get run twice
+//            } else if (TYPE(rbReturnValue) == T_STRING) {
+//                VALUE obj = rbffi_MemoryPointer_ForString(rbReturnValue);
+//
+//                *((void **) retval) = ((AbstractMemory *) DATA_PTR(obj))->address;
             } else {
                 *((void **) retval) = NULL;
             }
--- a/ext/ffi_c/Function.c
+++ b/ext/ffi_c/Function.c
@@ -844,10 +844,7 @@ invoke_callback(VALUE data)
             if (TYPE(rbReturnValue) == T_DATA) {
                 *((void **) retval) = ((AbstractMemory *) DATA_PTR(rbReturnValue))->address;
             } else if (TYPE(rbReturnValue) == T_STRING) {
-                VALUE s = StringValue(rbReturnValue);
-                VALUE args[] = { INT2FIX(1), LONG2NUM(RSTRING_LEN(s) + 1), Qfalse };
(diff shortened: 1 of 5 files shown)


ext/ffi_c/Function.c
```diff
@@ -839,6 +839,10 @@ invoke_callback(VALUE data)
             }
             break;
 
+        case NATIVE_STRING:
+            *((void **) retval) = ((AbstractMemory *) DATA_PTR(rbReturnValue))->address;
```
datibbaw: This fails horrible without `FFI::MemoryPointer.from_string` 😭

ext/ffi_c/Function.c
```diff
@@ -839,6 +840,21 @@ invoke_callback(VALUE data)
...
             break;
 
+        case NATIVE_STRING:
+            if (TYPE(rbReturnValue) == T_DATA) {
+                *((void **) retval) = ((AbstractMemory *) DATA_PTR(rbReturnValue))->address;
+            } else if (TYPE(rbReturnValue) == T_STRING) {
+                VALUE s = StringValue(rbReturnValue);
+                VALUE args[] = { INT2FIX(1), LONG2NUM(RSTRING_LEN(s) + 1), Qfalse };
+                VALUE obj = rb_class_new_instance(3, args, rbffi_MemoryPointerClass);
+                rb_funcall(obj, rb_intern("put_string"), 2, INT2FIX(0), s);
```
datibbaw: This would've been nicer if encapsulated as `rbffi_MemoryPointer_ForString`

larskanis:
Thank you for your contribution! I'm not the original author of the related code, but I'm pretty sure there is a reason why `:string` is not handled as a callback return value: It is not safe in most cases.

That's because the ruby String object that is returned by the block could get GC'ed as soon as the block has ended and so the string data can get invalid at this point in time. In fact the same is true for a `FFI::MemoryPointer.from_string(s)` but in that way you're getting aware about the fact, that you're returning a pointer. So you have to make sure, that the data the pointer is pointing to, is not GC'ed until it is processed by the C code. This can be solved by storing it in a variable outside of the block or in a instance variable or similar.

[JRuby doesn't accept](https://github.com/jruby/jruby/blob/4e8bb2666a7257f0f5986800f96bb88efdd6acbd/core/src/main/java/org/jruby/ext/ffi/jffi/NativeClosureProxy.java#L107) `:string` as a callback return value and raises a
```Ruby
TypeError: invalid callback return type: STRING
```
I think this check should be implemented here in C-FFI as well. I'm not convinced, that adding `:string` conversion to the callback return path is really a good idea, since it can lead to difficult to find GC related errors.

PR #755, opened 2020-03-25
Add TruffleRuby in CI and skip the few specs which are not yet supported on TruffleRuby

eregon:
Also adding a few extra specs and fixing the `long double` spec to actually run.
See the individual commit messages for details.

Part of https://github.com/ffi/ffi/issues/660#issuecomment-603895774


eregon:
`long double` specs seem to segfault on Windows, should we simply exclude them?
https://ci.appveyor.com/project/larskanis/ffi-aofqa/builds/31713641/job/p7uuncjd3t73838n

eregon:
I can split the long double change in another PR if you prefer (it's just a rename currently)

spec/ffi/long_double_spec.rb
```diff
@@ -6,13 +6,14 @@
...
-describe ":long_double arguments and return values" do
+# long double not yet supported on TruffleRuby
+describe ":long_double arguments and return values", :if => RUBY_ENGINE != "truffleruby" do
   module LibTest
     extend FFI::Library
     ffi_lib TestLibrary::PATH
     attach_function :add_f128, [ :long_double, :long_double ], :long_double
     attach_function :ret_f128, [ :long_double ], :long_double
-  end
+  end if RUBY_ENGINE != "truffleruby"
```
headius: Will this even run given the `:if` clause above?

spec/ffi/struct_spec.rb
```diff
@@ -671,7 +672,7 @@ class StructString < FFI::Struct
     attach_function :struct_s8s32_ret_s8s32, [ S8S32.by_value ], S8S32.by_value
     attach_function :struct_s32_ptr_s32_s8s32_ret_s32, [ :int, :pointer, :int, S8S32.by_value ], :int
     attach_function :struct_varargs_ret_struct_string, [ :int, :varargs ], StructString.by_value
-  end
+  end if RUBY_ENGINE != "truffleruby"
```
headius: Same issue here... doesn't the `:if` clause prevent this code from running?

spec/ffi/long_double_spec.rb
eregon: Yes, unfortunately this is needed (I though too it wouldn't be needed initially), and apparently RSpec's `if:` just skips the `it` examples but still yields to the block.

headius: 👍weird, seems like the wrong choice but 🤷‍♂️

larskanis:
LGTM, I'll merge this. It's probably better to separate the coerce test from the other struct tests. I'll open a PR for review.

eregon:
I'll rebase this to get the CI fixes.

There are still a few issues preventing the CI to pass on TruffleRuby, I'll keep working on it: https://travis-ci.org/github/ffi/ffi/builds/667891953

eregon:
@larskanis I see you merged it manually to master, thanks!
(except 1f6618f7c26073f1f83a218ee123f0c52e69d3d0, which doesn't matter)

larskanis:
Great work! I added f2680e692b7aa0cba31b63e3fa33e8c11e41c81d instead of 1f6618f7c26073f1f83a218ee123f0c52e69d3d0 . Could you please have a look at the remaining Truffleruby issues [here](https://travis-ci.org/github/ffi/ffi/builds/668207461)?

ext/ffi_c/LongDouble.c
```diff
@@ -41,7 +41,9 @@ rbffi_num2longdouble(VALUE value)
 
     if (RTEST(rb_cBigDecimal) && rb_cBigDecimal != rb_cObject && RTEST(rb_obj_is_kind_of(value, rb_cBigDecimal))) {
         VALUE s = rb_funcall(value, rb_intern("to_s"), 1, rb_str_new2("E"));
-        return strtold(RSTRING_PTR(s), NULL);
+        long double ret = strtold(RSTRING_PTR(s), NULL);
+        RB_GC_GUARD(s);
```
larskanis: Is this required in Truffleruby? In MRI there is no need for a guard here IMHO.

eregon: No it's not, TruffleRuby doesn't use the C extension for FFI.
I did it to try to debug this long double failure on Windows.
Why is it not needed on MRI, couldn't `s` be GC'd during the `strtold` call?

eregon:
> Could you please have a look at the remaining Truffleruby issues [here](https://travis-ci.org/github/ffi/ffi/builds/668207461)?

On macOS, it fails because TravisCI uses a too old version on macOS (10.13):
https://travis-ci.org/github/ffi/ffi/jobs/668207490
Moving those jobs to GitHub Actions would solve that (relates to #760), any thought on that? We could start moving the ones that are easy to move (MRI + TruffleRuby on Linux/macOS/could also move Windows), and leave the rest on TravisCI, that would speed up CI and test the main platforms on a more stable CI system.

[On Linux](https://travis-ci.org/github/ffi/ffi/jobs/668207484) I need to add the new LONG_DOUBLE_SIZE constant in TruffleRuby, since it was added in the C extension.
For the benchmark failure I'll need to investigate.

PR #756, opened 2020-03-25
Add checks and specs for writing to inline array and char array fields

eregon:
* Note that the last two of these specs fail on master, so I fixed the C extension to also have this behavior.

Part of https://github.com/ffi/ffi/issues/660#issuecomment-603895774

### DIFF
--- a/spec/ffi/struct_spec.rb
+++ b/spec/ffi/struct_spec.rb
@@ -1,3 +1,4 @@
+# -*- encoding: utf-8 -*-
 #
 # This file is part of ruby-ffi.
 # For licensing, see LICENSE.SPECS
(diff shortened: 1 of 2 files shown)


spec/ffi/struct_spec.rb
```diff
@@ -786,6 +786,62 @@ class StructWithArray < FFI::Struct
...
+    @s[:after] = -1
+    @s[:str] = "abcd"
+    expect(@s[:before]).to eq(0)
+    expect(@s[:str].to_s).to eq("abcd")
+    expect(@s[:after]).to eq(-1) # The above should not write to the next field
+  end
+
+  it 'raises when writing a longer String to the char array' do
+    expect { @s[:str] = "abcdefg" }.to raise_error(IndexError)
+  end
```
eregon: Do you agree these 2 specs should be the correct behavior?
Current output on MRI:
```
  1) FFI::Struct  with a char array field allows writing a String of the same size to the char array
     Failure/Error: expect(@s[:after]).to eq(-1) # The above should not write to the next field
     
       expected: -1
            got: -256
     
       (compared using ==)
     # ./spec/ffi/struct_spec.rb:839:in `block (2 levels) in <module:StructSpecs>'

  2) FFI::Struct  with a char array field raises when writing a longer String to the char array
     Failure/Error: expect { @s[:str] = "abcdefg" }.to raise_error(IndexError)
       expected IndexError but nothing was raised
     # ./spec/ffi/struct_spec.rb:843:in `block (2 levels) in <module:StructSpecs>'
```

eregon: I implemented that behavior in the C extension.

eregon:
Ready for review.

eregon:
I'll rebase this to get the CI fixes.

eregon:
@larskanis Could you review this when you get some time?
Should I add a ChangeLog entry or is that done at release time?

larskanis:
> Should I add a ChangeLog entry or is that done at release time?

Adding CHANGELOG entries as part of the commit tend to result in unnecessary merge conflicts. So I usually write the CHANGELOG just before the release and use this to sort it and pimp it up for better readability.

eregon:
Thanks, your added commit looks good.

Agreed, I think next minor version is fine, it would be a bug to rely on previous behavior (could overwrite unintended memory or lead to SEGV).

I'll let you merge the PR.

PR #758, opened 2020-03-27
Add typical SDK paths on macOS

larskanis:
macOS removed `/usr/include` in recent versions of macOS. This forces the usage of the internal libffi although macOS has a usable one itself.

Related to #757

### DIFF
--- a/ext/ffi_c/extconf.rb
+++ b/ext/ffi_c/extconf.rb
@@ -8,7 +8,11 @@ if !defined?(RUBY_ENGINE) || RUBY_ENGINE == 'ruby' || RUBY_ENGINE == 'rbx'
     # We need pkg_config or ffi.h
     libffi_ok = pkg_config("libffi") ||
         have_header("ffi.h") ||
-        find_header("ffi.h", "/usr/local/include", "/usr/include/ffi")
+        find_header("ffi.h",
+                    "/usr/local/include",
+                    "/usr/include/ffi",
+                    "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/ffi",
+                    "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/ffi")
 
     # Ensure we can link to ffi_call
     libffi_ok &&= have_library("ffi", "ffi_call", [ "ffi.h" ]) ||


ext/ffi_c/extconf.rb
```diff
@@ -8,7 +8,11 @@ def system_libffi_usable?
     # We need pkg_config or ffi.h
     libffi_ok = pkg_config("libffi") ||
         have_header("ffi.h") ||
-        find_header("ffi.h", "/usr/local/include", "/usr/include/ffi")
+        find_header("ffi.h",
+                    "/usr/local/include",
+                    "/usr/include/ffi",
+                    "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/ffi",
+                    "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/ffi")
```
eregon: Wouldn't using `xcrun --sdk macosx --show-sdk-path` to get the SDK path and then append `/usr/include/ffi` be more future-proof?

DomT4: Shelling out to `xcrun` can cause some unexpected behaviour unless you catch those cases and I'm not sure I'd recommend it.

Namely that if you haven't properly configured your development setup on macOS you'll get a graphical pop-up smack back in the middle of your screen asking if you'd like to install the CLT, and if you haven't accepted the Xcode/CLT license `xcrun` just errors out telling you to do that first. They are edge cases but, people love to stumble upon edge cases.

eregon: Compiling won't work until the CLT are installed anyway, so maybe it's fine to get the popup then if the user still did not install command-line tools?

Re the license, we could validate the output and only use it if it exists as a directory.

larskanis: I opened #765 for a version using `xcrun`. I don't have a strong opinion to any of both. And I don't have macOS, so maybe this is something for someone who can actually try the patch out :smile:

eregon:
Looks like macOS' FFI doesn't have `ffi_closure_alloc()`, so it's not used in TravisCI:
https://travis-ci.org/github/ffi/ffi/jobs/667797758?utm_medium=notification&utm_source=github_status

@larskanis Could you rebase so this also runs in GitHub Actions?
TravisCI macOS is pretty old.

eregon:
See https://github.com/ffi/ffi/pull/766#issuecomment-606232503, on GitHub Actions' macOS libffi seems recent enough.
It's found under `/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/ffi`, so I would suggest to just search that path and not bother with the Xcode MacOSX.sdk as that doesn't seem to have libffi.

Sorry for the comment about that one, seems the Xcode MacOSX.sdk isn't useful for libffi anyway.

ext/ffi_c/extconf.rb
```diff
@@ -8,7 +8,11 @@ def system_libffi_usable?
     # We need pkg_config or ffi.h
     libffi_ok = pkg_config("libffi") ||
         have_header("ffi.h") ||
-        find_header("ffi.h", "/usr/local/include", "/usr/include/ffi")
+        find_header("ffi.h",
+                    "/usr/local/include",
+                    "/usr/include/ffi",
+                    "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/ffi",
```
eregon: I suggest to simplify and just remove this path since it doesn't seem to have libffi.

eregon: Correction: that path has it too but seems less reliable and more likely to move in newer macOS versions.

PR #761, opened 2020-03-29
Fix remaining issues to make TruffleRuby pass the CI

eregon:
From https://github.com/ffi/ffi/pull/755#issuecomment-605619633

To fix the benchmarks we need fixes in TruffleRuby FFI's backend, so skip them for now.


eregon:
TravisCI doesn't even seem to start on this PR.

PR #763, opened 2020-03-30
Add ruby source files to the java gem

larskanis:
Related to https://github.com/ffi/ffi/pull/747 .

JRuby-9.3 is expected to be fully compatible to this gem's Ruby code. This allows to ship the Ruby library code per platform java gem and add it as a default gem to JRuby.

### DIFF
--- a/lib/ffi.rb
+++ b/lib/ffi.rb
@@ -8,7 +8,7 @@ if RUBY_ENGINE == 'ruby' || RUBY_ENGINE == 'rbx'
 
   require 'ffi/ffi'
 
-elsif RUBY_ENGINE == 'jruby'
+elsif RUBY_ENGINE == 'jruby' && Gem::Version.new(JRUBY_VERSION) >= Gem::Version.new("9.3.pre")
   JRuby::Util.load_ext("org.jruby.ext.ffi.FFIService")
   require 'ffi/ffi'
 
--- a/Rakefile
+++ b/Rakefile
@@ -65,16 +65,9 @@ task :default => :spec
 
 namespace 'java' do
 
-  java_gem_spec = Gem::Specification.new do |s|
-    s.name = gem_spec.name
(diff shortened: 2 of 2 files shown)


lib/ffi.rb
```diff
@@ -8,7 +8,7 @@
 
   require 'ffi/ffi'
 
-elsif RUBY_ENGINE == 'jruby'
+elsif RUBY_ENGINE == 'jruby' && Gem::Version.new(JRUBY_VERSION) >= Gem::Version.new("9.3.pre")
```
larskanis: @headius I'm not sure about pre 9.3 JRuby. Is this OK?

headius: @larskanis Pre-9.3 JRuby likely won't work **at all** with the FFI gem sources, so we won't want people to install it. I'm not sure if there's a good way to mitigate that, though. 🤔

Anyone on 9.2 that installs and activates the new version of the gem that includes sources might run into problems. I guess we just recommend they lock to an earlier FFI gem if they're not on 9.3+? Maybe a hard error here to make sure they know why it won't work?

larskanis: The above works pretty well for me on JRuby-9.2.9. It falls through to the next section
```ruby
# Remove the ffi gem dir from the load path, then reload the internal ffi implementation
```
and loads the builtin ffi implementation. I don't think a hard error or breaking compatibility is necessary.

headius: Aha, I see now. It's exotic but I guess that should work fine (and it's not new code).

godfat: @headius So I guess this is the reason for rspec-mocks test suites not working for JRuby 9.2.13.0: https://github.com/rspec/rspec-mocks/pull/1218#issuecomment-718944598

Not sure why it works for earlier JRuby though.

PR #765, opened 2020-03-30
Add SDK path on macOS

larskanis:
macOS removed `/usr/include` in recent versions of macOS. This forces the usage of the internal libffi although macOS has a usable one itself.

Related to #757 . This is an alternative approach to #758 .

### DIFF
--- a/ext/ffi_c/extconf.rb
+++ b/ext/ffi_c/extconf.rb
@@ -8,7 +8,8 @@ if RUBY_ENGINE == 'ruby' || RUBY_ENGINE == 'rbx'
     # We need pkg_config or ffi.h
     libffi_ok = pkg_config("libffi") ||
         have_header("ffi.h") ||
-        find_header("ffi.h", "/usr/local/include", "/usr/include/ffi")
+        find_header("ffi.h", "/usr/local/include", "/usr/include/ffi") ||
+        (find_header("ffi.h", `xcrun --sdk macosx --show-sdk-path`.strip + "/usr/include/ffi") rescue false)
 
     # Ensure we can link to ffi_call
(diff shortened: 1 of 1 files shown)


ext/ffi_c/extconf.rb
```diff
@@ -8,7 +8,8 @@ def system_libffi_usable?
     # We need pkg_config or ffi.h
     libffi_ok = pkg_config("libffi") ||
         have_header("ffi.h") ||
-        find_header("ffi.h", "/usr/local/include", "/usr/include/ffi")
+        find_header("ffi.h", "/usr/local/include", "/usr/include/ffi") ||
+        (find_header("ffi.h", `xcrun --sdk macosx --show-sdk-path`.strip) rescue false)
```
eregon: This needs + `/include/ffi`, otherwise it doesn't find it: https://github.com/ffi/ffi/pull/765/checks?check_run_id=546591203
`rescue false` seems brittle to me and might swallow unwanted exceptions.

eregon: I created https://github.com/ffi/ffi/pull/766 to show what I was thinking.

larskanis: > This needs + /include/ffi

Of course! I added "usr/include/ffi" and it seems to work.

To be honest I like this PR (including the rescue) more than #766 . It calls `xcrun` only when necessary, fits well into the chain, is very simple and doesn't break the installation if anything goes wrong.

eregon: Indeed, I forgot the `/usr` in front!
As you prefer, I'm thinking just searching `/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/ffi` as in #758 is the simplest now.

I personally dislike any silent `rescue`, as it can easily become a debugging nightmare.

PR #768, opened 2020-04-11
Use the Ruby files of the FFI gem on TruffleRuby >= 20.1.0

eregon:
Similar logic as for JRuby.
This will only apply for TruffleRuby >= 20.1.0 (not yet released), as versions before don't expect that.

I ran it locally with this and changes in TruffleRuby and all specs passed.

Ref #660

### DIFF
--- a/lib/ffi.rb
+++ b/lib/ffi.rb
@@ -8,8 +8,12 @@ if RUBY_ENGINE == 'ruby' || RUBY_ENGINE == 'rbx'
 
   require 'ffi/ffi'
 
-elsif RUBY_ENGINE == 'jruby' && Gem::Version.new(JRUBY_VERSION) >= Gem::Version.new("9.3.pre")
+elsif RUBY_ENGINE == 'jruby' && Gem::Version.new(RUBY_ENGINE_VERSION) >= Gem::Version.new("9.3.pre")
   JRuby::Util.load_ext("org.jruby.ext.ffi.FFIService")
+  require 'ffi/ffi'
+
+elsif RUBY_ENGINE == 'truffleruby' && Gem::Version.new(RUBY_ENGINE_VERSION) >= Gem::Version.new("20.1.0")
+  require 'truffleruby/ffi_backend'
   require 'ffi/ffi'
 
 else
--- a/lib/ffi.rb
+++ b/lib/ffi.rb
@@ -12,7 +12,7 @@ elsif RUBY_ENGINE == 'jruby' && Gem::Version.new(RUBY_ENGINE_VERSION) >= Gem::Ve
   JRuby::Util.load_ext("org.jruby.ext.ffi.FFIService")
   require 'ffi/ffi'
 
-elsif RUBY_ENGINE == 'truffleruby' && Gem::Version.new(RUBY_ENGINE_VERSION) >= Gem::Version.new("20.1.0")
+elsif RUBY_ENGINE == 'truffleruby' && Gem::Version.new(RUBY_ENGINE_VERSION) >= Gem::Version.new("20.1.0-dev-a")
   require 'truffleruby/ffi_backend'
   require 'ffi/ffi'
 
--- a/lib/ffi/pointer.rb
+++ b/lib/ffi/pointer.rb
@@ -31,17 +31,24 @@
 #
 
 require 'ffi/platform'
+
+# NOTE: all method definitions in this file are conditional on
+# whether they are not already defined. This is needed because
+# some Ruby implementations (e.g., TruffleRuby) might already
+# provide these methods due to using FFI internally, and we
+# should not override them to avoid warnings.
+
 module FFI
   class Pointer
 
     # Pointer size
-    SIZE = Platform::ADDRESS_SIZE / 8
+    SIZE = Platform::ADDRESS_SIZE / 8 unless const_defined?(:SIZE)
 
     # Return the size of a pointer on the current platform, in bytes
     # @return [Numeric]
     def self.size
       SIZE
-    end
+    end unless respond_to?(:size)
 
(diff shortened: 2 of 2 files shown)


lib/ffi/pointer.rb
```diff
@@ -31,17 +31,24 @@
 #
 
 require 'ffi/platform'
+
+# NOTE: all method definitions in this file are conditional on
+# whether they are not already defined. This is needed because
+# some Ruby implementations might already provide these methods
```
larskanis, line 37: Could you add that it means TruffleRuby specifically?

eregon: Done

eregon:
> Nice! I think we'll need TruffleRuby and JRuby in CI of this project.

`truffleruby-head` is already tested in GitHub Actions.
But indeed once TruffleRuby 20.1.0 is released we should test against `truffleruby` too to make sure the latest release works.

Regarding adding JRuby in CI I'll leave that to @headius or to you (it should be trivial to add it to `.github/workflows/ci.yml` but not sure it's green).

lib/ffi.rb
```diff
@@ -8,10 +8,14 @@
 
   require 'ffi/ffi'
 
-elsif RUBY_ENGINE == 'jruby' && Gem::Version.new(JRUBY_VERSION) >= Gem::Version.new("9.3.pre")
+elsif RUBY_ENGINE == 'jruby' && Gem::Version.new(RUBY_ENGINE_VERSION) >= Gem::Version.new("9.3.pre")
   JRuby::Util.load_ext("org.jruby.ext.ffi.FFIService")
   require 'ffi/ffi'
 
+elsif RUBY_ENGINE == 'truffleruby' && Gem::Version.new(RUBY_ENGINE_VERSION) >= Gem::Version.new("20.1.0")
```
eregon, line 15: We could tweak this condition to `>= Gem::Version.new("20.1.0-dev-a")` once the related changes are merged in TruffleRuby (nightly versions look like `20.1.0-dev-374d84d2`).
That way the CI running on truffleruby-head would return `true` for this condition.
(Potentially we could use `RUBY_RELEASE_DATE` too)

headius:
> Regarding adding JRuby in CI I'll leave that to @headius or to you (it should be trivial to add it to .github/workflows/ci.yml but not sure it's green).

Along with updating the FFI .rb files, I also made sure all specs are green on JRuby.

eregon:
@headius Could you make a PR adding `jruby-head` in CI then?
(Which has a version of `9.3.0.0-SNAPSHOT` which is >= `9.3.pre`)

headius:
@eregon I just checked and there's a handful of failures in recent spec updates related to the fixnum/bignum/long threshold.

I would prefer to wait on adding JRuby to CI until we can use the gem for all FFI ruby sources (i.e. delete our copies of libs and specs), so I know there's no conflict between FFI head and JRuby head.

eregon:
> I would prefer to wait on adding JRuby to CI until we can use the gem for all FFI ruby sources (i.e. delete our copies of libs and specs), so I know there's no conflict between FFI head and JRuby head.

That wouldn't work with older versions of the FFI gem though, so I'm sure how feasible is that.

lib/ffi.rb
eregon: I've done this, the latest TruffleRuby nightly now has the changes needed.

eregon:
@larskanis Looks like it's working just fine with `truffleruby-head`, OK to merge?

headius:
> That wouldn't work with older versions of the FFI gem though

I don't know what you mean here. Why would I test against an older version of the gem?

eregon:
> I don't know what you mean here. Why would I test against an older version of the gem?

Users might install an older version of the gem on the future JRuby release without FFI Ruby files. That would then break. So I think we need to keep Ruby files also in Ruby implementations as long as we want to support older FFI releases.

headius:
@eregon I think you misunderstand... we would still include FFI as a default gem, like most of the other standard libraries. Installing an older FFI gem would not install any new files, so the ones in stdlib would still be used. Newer FFI gems would be picked up and used. I don't see a problem.

headius:
FWIW I don't really intend to "support" older FFI gems in JRuby because they're all stub gems with no content.

eregon:
Right, that should work if that version of the default gem is recent or includes the JRuby version check in ffi.rb. A default gem is not very different than a copy in stdlib + a .gemspec anyway.

eregon:
@larskanis Could you make a release of FFI?

This change will help us address an issue happening with `aruba` + `bundle install --standalone` adding paths with `..` inside `$LOAD_PATH` (https://github.com/rubygems/rubygems/issues/3505), making https://github.com/ffi/ffi/blob/9e6892c40c5751c82ccf8a4c45f53b78f22824c3/lib/ffi.rb#L21 not working as expected.
We could also try to adapt the code there to compare against expanded (`File.expand_path`) entries of `$LOAD_PATH` but I think that would be kind of slow and I think it's mostly a Bundler issue. Also, with this PR, TruffleRuby no longer uses that part of the code anyway, which is cleaner and avoids the "require gem but remove it and go to stdlib" part.

headius:
> A default gem is not very different than a copy in stdlib + a .gemspec anyway.

Isn't that exactly what a default gem is?

eregon:
Yeah, exactly so in the end whether there is a copy of the `lib` files in JRuby or a default gem is pretty much the same, that's why I mentioned that.

larskanis:
@eregon 

> Could you make a release of FFI?

I would like to fix #751 first, so that a `:string` callback return value is properly raised. The rest is fine for ffi-1.13.0 IMHO. I'll release it in the next days.

PR #770, opened 2020-04-16
Implement callback returning type :string

larskanis:
Callbacks returning a `:string` type were not supported so far. It was possible to define such a callback, but the value returned was `NULL` in any case. This implementation follows the semantics for passing `:string` parameters to functions as [implemented in `Call.c`](https://github.com/ffi/ffi/blob/31db4122f7ee9b67323b3e28c00b35e8d374a3cd/ext/ffi_c/Call.c#L296-L306). That means in particular usage of `StringValueCStr()`, so that no zero bytes are allowed in the string.

Special care has to be taken regarding lifetime of the returned text pointer. The C pointer to the string is only valid until the next GC triggering function is called, unless the String object is hold in Ruby outside of the callback block. Even when the object is stored outside of the Ruby block, the text pointer can move when GC.compact is called. So `:string` returning callbacks can safely be used only, when the referenced text is processed immediately by the calling C function.

Return type `:string` is already implemented in TruffleRuby, but [fails in JRuby](https://github.com/jruby/jruby/blob/4e8bb2666a7257f0f5986800f96bb88efdd6acbd/core/src/main/java/org/jruby/ext/ffi/jffi/NativeClosureProxy.java#L107) and raises a
```Ruby
TypeError: invalid callback return type: STRING
```

Resolves #751

- [ ] Update the Wiki accordingly

@eregon, @headius  What do you think about this addition? Should we add `:string` returning callbacks in JRuby or deny them on all implementations?

### DIFF
--- a/ext/ffi_c/Function.c
+++ b/ext/ffi_c/Function.c
@@ -793,6 +793,15 @@ invoke_callback(VALUE data)
         case NATIVE_FLOAT64:
             *((double *) retval) = NUM2DBL(rbReturnValue);
             break;
+
+        case NATIVE_STRING:
+            if (NIL_P(rbReturnValue)) {
+                *((void **) retval) = NULL;
+            } else {
+                *((void **) retval) = StringValueCStr(rbReturnValue);
+            }
+            break;
+
         case NATIVE_POINTER:
             if (TYPE(rbReturnValue) == T_DATA && rb_obj_is_kind_of(rbReturnValue, rbffi_PointerClass)) {
                 *((void **) retval) = ((AbstractMemory *) DATA_PTR(rbReturnValue))->address;
--- a/spec/ffi/callback_spec.rb
+++ b/spec/ffi/callback_spec.rb
@@ -51,6 +51,7 @@ module CallbackSpecs
       callback :cbVrUL, [ ], :ulong
       callback :cbVrS64, [ ], :long_long
       callback :cbVrU64, [ ], :ulong_long
+      callback :cbVrA, [], :string
       callback :cbVrP, [], :pointer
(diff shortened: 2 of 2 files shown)


larskanis:
@eregon How long is a `char*` string returned by a callback alive in TruffleRuby? To what point in time must the string pointer be considered as invalid? Could there be any relocation even when the string is still bound in Ruby space (similar to `GC.compact`).

eregon:
@larskanis It's similar to what you described above, as soon as the Ruby String is no longer referenced then the native `char*` pointer can be freed.

The `char*` pointer itself never moves on TruffleRuby (basically a `malloc()`) as long as the String is not reallocated (by appending or removing characters from the String).
I think on MRI it's also malloc() if the string is long enough (> 23 chars), but if not it's embedded directly in the `struct RString` and then I guess it can be moved along the Ruby String object by `GC.compact` as you said above.

eregon:
Actually, the current semantics on TruffleRuby are different than I thought.
@rschatz who wrote Truffle NFI clarified that.
For a callback returning `:string`, `strdup()` will be called on it, which means the C code receiving the `char*` should then `free()` it.
This avoids any lifetime issue, but requires the caller to `free()` it.

larskanis:
Closing this in favor of #782 .

The validity of the `char*` returned by a callback with string type is not well enough defined. It also seems dependent on the ruby implementation. Therefore we now reject `:string` return type for callbacks.

@eregon Thank you for investigating the Truffleruby conversion of strings! I also inspected the Truffleruby's FFI code a bit, but wasn't sure about how and when the memory is freed.

PR #775, opened 2020-05-10
Add a types.conf for powerpc-openbsd. Tested on a G4 7455

jsfierro:
This adds support for PowerPC OpenBSD systems. Generated on a G4 7455 and it works fine.


larskanis:
Looks good! Thank you!

Issue #776, opened 2020-05-15
FFI::Struct#layout fails when mkmf is required


eregon:
```
$ ruby -rmkmf -rffi -e 'module A; class A < FFI::Struct; layout :a, :int; end; end' |& cat
/home/eregon/.rubies/ruby-2.6.6/lib/ruby/2.6.0/mkmf.rb:1255:in `find_type': wrong number of arguments (given 1, expected 2+) (ArgumentError)
	from /home/eregon/.rubies/ruby-2.6.6/lib/ruby/gems/2.6.0/gems/ffi-1.12.2/lib/ffi/struct.rb:271:in `find_type'
	from /home/eregon/.rubies/ruby-2.6.6/lib/ruby/gems/2.6.0/gems/ffi-1.12.2/lib/ffi/struct.rb:265:in `find_field_type'
	from /home/eregon/.rubies/ruby-2.6.6/lib/ruby/gems/2.6.0/gems/ffi-1.12.2/lib/ffi/struct.rb:305:in `array_layout'
	from /home/eregon/.rubies/ruby-2.6.6/lib/ruby/gems/2.6.0/gems/ffi-1.12.2/lib/ffi/struct.rb:217:in `layout'
	from -e:1:in `<class:A>'
	from -e:1:in `<module:A>'
	from -e:1:in `<main>'
```

This is due to https://github.com/ffi/ffi/blob/33fb45a80e20fd9d9c39de3753abca9b4ef3abe7/lib/ffi/struct.rb#L247
and `mkmf` defining a public `find_type` on every object by `include MakeMakefile` :disappointed: 

On TruffleRuby, the socket stdlib uses FFI, so this bug ends up failing to install some gems that require `socket` (indirectly) in `extconf.rb`, see https://github.com/oracle/truffleruby/issues/1896

eregon:
I filed an issue for mkmf too since it doesn't reasonable to have so many public methods:
https://bugs.ruby-lang.org/issues/16896

But we'll need a workaround in FFI, as old releases won't be fixed.

eregon:
Code there is related to 510c99c54b933c60ece0d1c01291db17f01a5c65/#256 and b49cfae8eb456230c680d22fc25e8bc4eac6ef3f.
It's not clear why FFI ever checked for `respond_to?(:find_type)` but that seems quite brittle as this bug shows.

eregon:
I think the actual intention was to restrict to FFI::Library or FFI::Struct, so the condition should be:
[...]

PR #777, opened 2020-05-15
Improve check to detect if a module has a #find_type method suitable for FFI

eregon:
* Fixes https://github.com/ffi/ffi/issues/776


larskanis:
Makes sense! Thank you!

PR #779, opened 2020-05-18
MSVC 2019 Fixes

cfis:
These 5 commits fix various issues that prevent Microsoft Visual Studio 2019 from compiling the ffi bindings. Each commit provides details about the specific issue it is fixing.

### DIFF
--- a/ext/ffi_c/Buffer.c
+++ b/ext/ffi_c/Buffer.c
@@ -114,7 +114,7 @@ buffer_initialize(int argc, VALUE* argv, VALUE self)
         }
 
         /* ensure the memory is aligned on at least a 8 byte boundary */
-        p->memory.address = (void *) (((uintptr_t) p->data.storage + 0x7) & (uintptr_t) ~0x7UL);
+        p->memory.address = (void *) (((uintptr_t) p->data.storage + 0x7) & (uintptr_t) ~0x7ULL);
     
         if (p->memory.size > 0 && (nargs < 3 || RTEST(rbClear))) {
             memset(p->memory.address, 0, p->memory.size);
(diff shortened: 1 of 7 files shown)


larskanis:
@cfis Thank you for your contributions!

There are 3 things I worry about:

1. `ucrtbase.dll` is not fully compatible to `msvcrt.dll`. You might noticed the broken CI run https://github.com/ffi/ffi/pull/779/checks?check_run_id=686393082#step:9:1339 , which is caused by `time()` not being defined as a symbol. That was [easy to fix](https://github.com/ffi/ffi/commit/911d103be24b299c1c77f6516e20eff815cb0e0d) but it proves to be not fully backward compatible. So far I didn't find any real world incompatibilities in other projects using FFI.
2. `ucrtbase.dll` is not specified by MS to be compatible to anything - only the lib files are mentioned in CRT docs. I also found [this article](https://mingwpy.github.io/ucrt.html) which states: 

> it’s actually forbidden to link to ucrtbase.dll directly and one must use the official .lib file, or else one loses all compatibility guarantees.

3. There are currently no CI runs for ffi on MSVC, so that the compatibility can easily break again.

cfis:
Thanks for merging. I agree with your points.

1. Good question about time - as you noticed it is not exported by ucrtbase (which does have _time_32 and _time_64).  The MSFT docs say time is a "wrapper" around those 2 functions. Its not hard to work around if you need to access time (or whatever else is missing), just in the additional library to your fii code. But I agree its unexpected and annoying. Sigh.

2. Interesting article. It links to this page - https://gist.github.com/njsmith/08b1e52b65ea90427bfd which list "virtual" dll names. And that got me to https://stackoverflow.com/questions/47529106/what-are-api-ms-win-lx-x-x-dll-umbrella-libraries. Which is an explanation of how windows now has "virtual" dll names that map to logical dlls. Which could be another approach I guess of how to get at the runtime library.

3. Yeah, that's definitely true. But since I'm the only one who seems to have noticed, its probably not all that important. My use case for using an MSVC ruby version is that MSVC provides, in my opinion of course, a much better debugging environment for working with C/C++ code making easier to work on Ruby native extensions I maintain.

PR #786, opened 2020-06-02
Implement MemoryPointer.from_string in Ruby instead of C

larskanis:
This is more readable and should be faster due to inline caching.


larskanis:
Little benchmark turned out, that this PR makes `MemoryPointer#from_string` approx. 10 to 15% slower.

Without `respond_to?` and `to_str` (which corresponds to `StringValue(to_str)` in C) the method is minimal faster. But it doesn't raise a proper TypeError and doesn't do implicit conversion to string.

PR #794, opened 2020-06-26
provide lib/ffi/platform/aarch64-openbsd/types.conf

buzzdeee:
This is basically a clone of the x86_64-openbsd/types.conf, I verified all __ types that they are the same on both platforms, to be pretty sure, the rest is as well.


larskanis:
LGTM! Thank you!

Issue #800, opened 2020-07-08
SEVG on Apple silicon macOS


Watson1978:
I tried sample code in README on Apple silicon macOS, it causes SEGV.

```
$ ruby -v
ruby 2.8.0dev (2020-07-08T14:27:10Z master 9fc25811d5) [arm64-darwin20]

$ gem i ffi
Fetching ffi-1.13.1.gem
Building native extensions. This could take a while...
Successfully installed ffi-1.13.1
1 gem installed


$ ruby ffi.rb
/Users/watson/.rbenv/versions/2.8.0/lib/ruby/gems/2.8.0/gems/ffi-1.13.1/lib/ffi/library.rb:275: [BUG] Bus Error at 0x0000000103460000
ruby 2.8.0dev (2020-07-08T14:27:10Z master 9fc25811d5) [arm64-darwin20]

-- Crash Report log information --------------------------------------------
   See Crash Report log file under the one of following:                    
     * ~/Library/Logs/DiagnosticReports                                     
     * /Library/Logs/DiagnosticReports                                      
   for more details.                                                        
Don't forget to include the above Crash Report log file in bug reports.     

-- Control frame information -----------------------------------------------
c:0005 p:---- s:0035 e:000034 CFUNC  :attach
c:0004 p:0250 s:0029 e:000028 METHOD /Users/watson/.rbenv/versions/2.8.0/lib/ruby/gems/2.8.0/gems/ffi-1.13.1/lib/ffi/library.rb:275
c:0003 p:0032 s:0009 e:000008 CLASS  ffi.rb:6
c:0002 p:0013 s:0006 e:000005 EVAL   ffi.rb:3 [FINISH]
c:0001 p:0000 s:0003 E:0021f0 (none) [FINISH]

-- Ruby level backtrace information ----------------------------------------
ffi.rb:3:in `<main>'
ffi.rb:6:in `<module:MyLib>'
/Users/watson/.rbenv/versions/2.8.0/lib/ruby/gems/2.8.0/gems/ffi-1.13.1/lib/ffi/library.rb:275:in `attach_function'
/Users/watson/.rbenv/versions/2.8.0/lib/ruby/gems/2.8.0/gems/ffi-1.13.1/lib/ffi/library.rb:275:in `attach'

-- Other runtime information -----------------------------------------------

... snip ...
```

Full crash log is in https://gist.github.com/Watson1978/85f1b5ff076a2d3e36057af0ab854299

### Test code
```
require 'ffi'

[...]

PR #801, opened 2020-07-08
Add types.conf of Apple silicon mac

Watson1978:
I just run following command on Apple silicon Mac

```
$ system_profiler SPHardwareDataType
Hardware:

    Hardware Overview:

      Model Name: Apple Development Platform
      Model Identifier: ADP3,2
      Processor Name: Apple A12Z Bionic
      Number of CPUs: 8
      L2 Cache (per CPU): 8 MB
      Memory: 16 GB
      Serial Number (system): C07CR95UPYPQ
      Hardware UUID: 39211857-BD84-57C6-B060-DEBF788CC50F
      Activation Lock Status: Disabled

$ rake types_conf

```

Related to https://github.com/ffi/ffi/issues/800

Thanks


larskanis:
Thank you for providing the type list on this new platform! I think the architecture and the path should be called "aarch64" instead of "arm". I opened #802 to fix this. If you agree, could you please verify that the `rake types_conf` generates the file in `aarch64-darwin/types.conf`?

Watson1978:
I tried #802, however, it still generate `arm-darwin` directory.

```
[pr/802][~/prj/ffi]$ rake types_conf
[pr/802][~/prj/ffi]$ git status
On branch pr/802
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        lib/ffi/platform/arm-darwin/

nothing added to commit but untracked files present (use "git add" to track)
```

Seems that `RbConfig::CONFIG['host_cpu']` has `arm` value.
https://gist.github.com/Watson1978/16c3f8b2673a8db266b2332b73cb1bc9

I'm wondering whether it's a Ruby problem

Watson1978:
I'm confirming this behavior in https://bugs.ruby-lang.org/issues/17021

headius:
What is the status of this? As I am sure @larskanis knows, JRuby's FFI layer also depends on libffi, and we are eagerly awaiting a usable build.

larskanis:
This is more or less blocked by https://bugs.ruby-lang.org/issues/17021 . We could however add a workaround until the RbConfig variables are fixed.

eregon:
I'd recommend to workaround for now, and assume any `^arm` on `darwin` is `aarch64`.
It would be good to fix the bug upstream, but even if it is fixed it won't be applied to existing Ruby releases.

eregon:
@larskanis The bug was fixed upstream.
Nobu recommends to use `target_cpu`: https://bugs.ruby-lang.org/issues/17021#note-8
I think we should try that.

larskanis:
I tried using `target_cpu` instead of `host_cpu` in [this branch](https://github.com/larskanis/ffi/tree/arm-darwin). However I'm uncertain if it breaks too much. For instance:

  * 32 bit Mingw uses "target_cpu"=>"i386" and "host_cpu"=>"i686"
  * 64 bit Minwg uses "target_cpu"=>"x64" and "host_cpu"=>"x86_64"

While `i386` vs. `i686` is properly handled in [platform.rb](https://github.com/ffi/ffi/blob/187328eb3c9e3088e2e3704e7ca68ae9928dee44/lib/ffi/platform.rb#L66), the term `x64` is not recognized.

I posted https://github.com/ffi/ffi/pull/843 which still uses `host_cpu` but adds a workaround for MacOS.

eregon:
Agreed, I remember also using `target_os` sometimes gives different values.

BTW, that Regexp `/i?86|x86|i86pc/` seems buggy, it should probably be `/i\d?86|x86|i86pc/`

PR #806, opened 2020-07-11
Fix FFI::Pointer#write_string to always terminate with a \0 byte

### LINKED ISSUE #805: #write_string is inconsistent with #put_string and does not NUL-terminates
For I think all `FFI::Pointer` methods, `put_*` is the same as `write_*` + an extra offset argument (e.g., `write_int(value)` and `put_int(offset, value)`). Same for `get_*` and `read_*`. Also for `*_array_of_*` variants.

Or in other words one can implement `write_foo(*args)` with `put_foo(0, *args)` (or the [other way around](https://github.com/oracle/truffleruby/blob/d6e56aa89fdf512e2f0b1b69bd54cffef114c047/src/main/ruby/truffleruby/core/truffle/ffi/pointer_access.rb#L289-L292)).

`read_string(length = nil)` and `get_string(offset, length = nil)` follow this pattern too, although they have an extra `length` argument.

But unfortunately `write_string` and `put_string` have a very important difference, creating an inconsistency:
https://github.com/ffi/ffi/blob/aa7db16ba59b005edb853b2791d9c43577419f6a/lib/ffi/pointer.rb#L101-L105
https://github.com/ffi/ffi/blob/d9fe6c703bfa6866b628549d12a43d489e53665c/ext/ffi_c/AbstractMemory.c#L508-L521
`write_string` does not NUL-terminates (surprising!) but `put_string` does.
```ruby
[9] pry(main)> p=FFI::MemoryPointer.new(8); p.write_int64(-1); p.write_string "abc"; p.read_bytes(8)
=> "abc\xFF\xFF\xFF\xFF\xFF"
[10] pry(main)> p=FFI::MemoryPointer.new(8); p.write_int64(-1); p.put_string 0, "abc"; p.read_bytes(8)
=> "abc\x00\xFF\xFF\xFF\xFF"
[11] pry(main)> p=FFI::MemoryPointer.new(8); p.write_int64(-1); p.write_string "abc"; p.read_bytes(8); p.read_string
=> "abc\xFF\xFF\xFF\xFF\xFF"
[12] pry(main)> p=FFI::MemoryPointer.new(8); p.write_int64(-1); p.put_string 0, "abc"; p.read_bytes(8); p.read_string
=> "abc"
```
A successive `read_string` doesn't work as expected, so that's another inconsistency, this time between `write_string` and `read_string`.

And it's been like that for a while, at least since https://github.com/ffi/ffi/commit/0bfdb2d5283da73a20c8136d2008c26bf71138d7#diff-430e60c6d45ffe238bb858feb0605e97R96

In fact, `write_string` behaves like `write_bytes`, but I think it should behave like `put_string` and NUL-terminates as well.

I think we should fix `write_string` to NUL-terminate.
I'm not sure regarding compatibility though.
@larskanis @headius Thoughts?

eregon:
* Add specs for FFI::Pointer#write_string and #put_string
* Fixes #805

There was no specs for `write_string` yet, so actually changing as proposed in #805 caused no spec failures.

### DIFF
--- a/lib/ffi/pointer.rb
+++ b/lib/ffi/pointer.rb
@@ -97,11 +97,15 @@ module FFI
     # @param [Numeric] len length of string to return
     # @return [self]
     # Write +str+ in pointer's contents, or first +len+ bytes if 
-    # +len+ is not +nil+.
+    # +len+ is not +nil+. A final \0 byte is written if +len+ is not given.
     def write_string(str, len=nil)
-      len = str.bytesize unless len
-      # Write the string data without NUL termination
-      put_bytes(0, str, 0, len)
+      if len
+        # No NUL termination
+        put_bytes(0, str, 0, len)
+      else
+        # With NUL termination
+        put_string(0, str)
+      end
     end unless method_defined?(:write_string)
 
     # @param [Type] type type of data to read from pointer's contents
--- a/lib/ffi/pointer.rb
+++ b/lib/ffi/pointer.rb
@@ -102,7 +102,7 @@ module FFI
     def write_string(str, len=nil)
       if len
         if len == size
-          warn "[DEPRECATION] Memory to small to write a final 0-byte in #{caller[0]}. This will raise an error in ffi-2.0. Please use write_bytes instead or enlarge the memory region."
+          warn "[DEPRECATION] Memory too small to write a final 0-byte in #{caller(1, 1)[0]}. This will raise an error in ffi-2.0. Please use write_bytes instead or enlarge the memory region."
           put_bytes(0, str, 0, len)
         else
           put_char(len, 0) # Check size before writing str
@@ -110,7 +110,7 @@ module FFI
         end
       else
         if str.bytesize == size
-          warn "[DEPRECATION] Memory to small to write a final 0-byte in #{caller[0]}. This will raise an error in ffi-2.0. Please use write_bytes instead or enlarge the memory region."
+          warn "[DEPRECATION] Memory too small to write a final 0-byte in #{caller(1, 1)[0]}. This will raise an error in ffi-2.0. Please use write_bytes instead or enlarge the memory region."
           put_bytes(0, str)
         else
           put_string(0, str)
--- a/lib/ffi/pointer.rb
+++ b/lib/ffi/pointer.rb
@@ -85,12 +85,12 @@ module FFI
     # @param [String] str string to write
     # @param [Numeric] len length of string to return
     # @return [self]
-    # Write +len+ first bytes of +str+ in pointer's contents.
+    # Write +len+ first bytes of +str+ in pointer's contents and a final \0 byte.
     #
     # Same as:
     #  ptr.write_string(str, len)   # with len not nil
     def write_string_length(str, len)
-      write_bytes(str, 0, len)
+      write_string(str, len)
     end unless method_defined?(:write_string_length)
 
     # @param [String] str string to write
--- a/lib/ffi/pointer.rb
+++ b/lib/ffi/pointer.rb
@@ -101,10 +101,20 @@ module FFI
     # In both cases a final \0 byte is written after the string.
     def write_string(str, len=nil)
       if len
-        put_bytes(0, str, 0, len)
-        put_char(len, 0)
+        if len == size
+          warn "[DEPRECATION] Memory to small to write a final 0-byte in #{caller[0]}. This will raise an error in ffi-2.0. Please use write_bytes instead or enlarge the memory region."
+          put_bytes(0, str, 0, len)
+        else
+          put_char(len, 0) # Check size before writing str
+          put_bytes(0, str, 0, len)
+        end
       else
-        put_string(0, str)
+        if str.bytesize == size
+          warn "[DEPRECATION] Memory to small to write a final 0-byte in #{caller[0]}. This will raise an error in ffi-2.0. Please use write_bytes instead or enlarge the memory region."
+          put_bytes(0, str)
+        else
+          put_string(0, str)
+        end
       end
     end unless method_defined?(:write_string)
 
--- a/lib/ffi/pointer.rb
+++ b/lib/ffi/pointer.rb
@@ -96,14 +96,14 @@ module FFI
     # @param [String] str string to write
     # @param [Numeric] len length of string to return
     # @return [self]
-    # Write +str+ in pointer's contents, or first +len+ bytes if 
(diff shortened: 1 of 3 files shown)


eregon:
@larskanis Could you review?

The CI fails on TruffleRuby as expected (If we are OK with this change I'll apply it in TruffleRuby before merging) but it's otherwise passing.

eregon:
> I'm in a struggle with mixing null termination and no null termination. IMHO `write_string` should always add a null termination.

I kind of agree, and `write_bytes` can be used for the behavior without null termination.
I'm a little worried that people might expect `ptr.write_string("str", 3)` to write 3 and not 4 bytes though.

I'll update the PR to always add a `\0` for `write_string`.

larskanis:
@eregon I changed code and specs to add a null character in any case in `write_string`. The Truffleruby tests fail now as expected. Can you adjust it as well?

lib/ffi/pointer.rb
```diff
@@ -96,12 +96,16 @@ def write_string_length(str, len)
...
+    # Write +str+ in pointer's contents.
+    # If +len+ is given, write the first +len+ bytes of +str+
+    # and append a final \0 byte after the string.
     def write_string(str, len=nil)
-      len = str.bytesize unless len
-      # Write the string data without NUL termination
-      put_bytes(0, str, 0, len)
+      if len
+        put_bytes(0, str, 0, len)
+        put_bytes(len, "\0")
```
eregon, line 105: I think we can use `put_char(len, 0)` here instead, which should be a bit more efficient.

eregon: Done

eregon:
I did the change in TruffleRuby, will ping here when it's merged and there is a new truffleruby-head build.

eregon:
@larskanis I'm not sure if we can do this change regarding compatibility.

In TruffleRuby's own usages of `write_string`, *all* [16 of them](https://gist.github.com/eregon/6772d6435bcd489b983cde2f6b9bd8cd) actually expected no final `\0`.
A good part of that code is originally from Rubinius, which clearly had the semantics of not appending a final `\0` for `write_string` (OTOH Rubinius FFI had other incompatibilities, some quite unintuitive).
All of them were either single-argument or `write_string(str, str.bytesize)` (so still writes the full String).
I'll change them to `write_bytes(str)`.
Migrating to `write_bytes` is not necessarily trivial, because `write_string(str, str.bytesize)` becomes `write_bytes(str)`.
`write_bytes(str, str.bytesize)` would start writing at offset `str.bytesize` which is not what we want.
So if we write only part of the string then `write_string(str, len)` => `write_bytes(str, 0, len)`.

I also noticed a problematic usage in OptCarrot:
https://github.com/mame/optcarrot/blob/7a38ff7ed8121fdd890bec201f98ebb3e750c611/lib/optcarrot/driver/misc.rb#L120

When running on CRuby there is at least a nice error:
```
/home/eregon/.rubies/ruby-2.6.6/lib/ruby/gems/2.6.0/gems/ffi-1.13.1/lib/ffi/pointer.rb:107:in `put_string': Memory access offset=0 size=1025 is out of bounds (IndexError)
	from /home/eregon/.rubies/ruby-2.6.6/lib/ruby/gems/2.6.0/gems/ffi-1.13.1/lib/ffi/pointer.rb:107:in `write_string'
	from /home/eregon/code/truffleruby-ws/truffleruby/bench/optcarrot/lib/optcarrot/driver/misc.rb:109:in `icon_data'
```
So this should help migration, but I'm not sure if the Pointer's size is always known.

On TruffleRuby there are currently no such checks, and so it results in malloc() assertions & errors.
So on the plus side, this really motivates me to implement the bound checks for FFI pointers on TruffleRuby :D

eregon:
@larskanis I'm OK to merge the change in TruffleRuby-head to experiment.
But if we do so, I'd like to have a FFI release soon, so any issue related to this change gets discovered quickly and the TruffleRuby FFI has the same behavior as the last FFI release.
Sounds good?

eregon:
@larskanis Could you reply here, do you think we should try this change and see what happens?
I could merge the change in TruffleRuby soon, TruffleRuby is starting a new development cycle. Then I think we should have a FFI release soon after to get feedback whether it's OK or too incompatible.

larskanis:
Thank you Benoit for your investigation [above](https://github.com/ffi/ffi/pull/806#issuecomment-698357515)! I'm sorry that I've got very little time for OSS development over the past months.

Regarding this change I'm unsure how to solve it best. Obviously we break compatibility in some cases, but we shouldn't in a minor version. We could use `Pointer#size` to check if we have enough space to write the string with termination. We then have two options:
1. Refine this PR: If the size permits, then write the string with termination. If the size doesn't permit, write the string without the termination and print a deprecation warning. The size check and the deprecation warning can be removed with ffi-2.0.
2. Reject this PR: Don't write the string with termination. If the size wouldn't allow termination, then print a deprecation warning instead. The string termination will be introduced with ffi-2.0.

I'm somewhat in favor with option 1, but still unsure. If we reach an agreement on this PR, I can release a new ffi version 1.14.0 within some days.

larskanis:
@eregon I added a commit for option 1 [above](https://github.com/ffi/ffi/pull/806#issuecomment-718224520). What do you think?

eregon:
@larskanis That sounds good to me.

In the case the pointer is unbounded (size=LONG_MAX), then an extra `\0` will be written, but I think there is no way to warn for that (the only alternative would be to delay that change until FFI 2).

Given there is a deprecation warning, maybe we could actually use the new behavior after a few FFI 1.x releases?
I think requiring a new major version for this seems a bit much to me.
It's more like a Ruby 2.5->2.6 change, not Ruby 2->3 change (but Ruby versioning is weird).

eregon:
What should we do about `FFI::Pointer#write_string_length`?
Same as `write_string` or same as `write_bytes`?

Because of the name (`string` and not `bytes`) it seems better to just call `write_string(str, len)` and \0-terminate.
Then we'll need to update the docs.

Currently the documentation is a mix:
```ruby
    # @param [String] str string to write
    # @param [Numeric] len length of string to return
    # @return [self]
    # Write +len+ first bytes of +str+ in pointer's contents.
    #
    # Same as:
    #  ptr.write_string(str, len)   # with len not nil
    def write_string_length(str, len)
      write_bytes(str, 0, len)
    end
```

larskanis:
I think it's good to merge now. There are some errors on Truffleruby, since it's not implemented there. Can I merge the PR now or is there anything else to consider?

eregon:
Fine to merge for me, I should be able to fix TruffleRuby soon after.

PR #810, opened 2020-07-17
Add specs specific to Centos callback bug #621

larskanis:
I forgot about these specs when fixing #621. They should be merged, to verify this use case in the future.

### DIFF
--- a/spec/ffi/fork_spec.rb
+++ b/spec/ffi/fork_spec.rb
@@ -1,0 +1,41 @@
+#
+# This file is part of ruby-ffi.
+# For licensing, see LICENSE.SPECS
+#
+
+require File.expand_path(File.join(File.dirname(__FILE__), "spec_helper"))
+
+# Skip platforms that don't implement Process.fork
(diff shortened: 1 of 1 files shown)


spec/ffi/fork_spec.rb
```diff
@@ -0,0 +1,41 @@
+#
+# This file is part of ruby-ffi.
+# For licensing, see LICENSE.SPECS
+#
+
+require File.expand_path(File.join(File.dirname(__FILE__), "spec_helper"))
+
+# Skip platforms that don't implement Process.fork
+unless RUBY_ENGINE == "truffleruby" or FFI::Platform.windows?
```
eregon, line 9: A more reliable way to test that would be `Process.respond_to?(:fork)`
(https://github.com/oracle/truffleruby/blob/master/doc/user/compatibility.md#fork)
I'll push a commit to do that.

eregon: 522581c9d43544ec36d13a761ed43f5b6b891119

larskanis: Thanks, seems to work!

PR #811, opened 2020-07-17
Avoid relocation of strings by moving the content to heap memory

larskanis:
This is due to the fact that Ruby-2.7+ relocates emdedded strings by `GC.compact`. See https://bugs.ruby-lang.org/issues/17023 for more description of the issue.

Unfortunately there's no good way to pin string values to a fixed memory location in the ruby C API. The only way seems to be `rb_gc_register_address()`, but this function would keep a reference to the string object forever, so that it will never be freed.

To work around this issue the string capacity is expanded, so that the string content is moved from embedded string memory to ordinary heap memory. The underlying string buffer of such strings is never relocated at `GC.compact` .

There are two side effects:

1. All strings smaller than 24 characters are copied when passed to the C function.
   This decreases the performance although this copying isn't required in almost all cases.
   It's only needed when `GC.compact` is used and the string in question is accessed after `GC.compact`.

2. The string object is modified in a way that changes it's `RSTRING_PTR()`.
   That isn't visible in ruby, but could be noticed in conjunction with other C extensions.


uvlad7:
Is the string isn't used in C after the function returns, it shouldn't be affected, right? AFAIK GC won't move objects on the stack.
So I'd suggest to change the contract for strings passed to C and used after the function returns (seems to be a pretty rare case) and say thay should not only be alive, but also pinned, and introduce `pin` helper which will basically do
```ruby
def pin(str)
  str.instance_variable_set(:@_ffi_pin, Fiddle::Pinned.new(str))
end
```
(ofc use FFI's own `Fiddle::Pinned` implementation and probably set an instance var without the `@` that can't be accessed in Ruby)
`unpin` method probably isn't needed - the "the string isn't used in C anymore but I need to keep in in Ruby and want to make it movable again" use case seems to be unrealistic

PR #815, opened 2020-08-03
Fix memory leak in MethodHandle

Watson1978:
`handle` has memory area allocated in `rbffi_MethodHandle_Alloc()`

https://github.com/ffi/ffi/blob/faae322f74969dfd9a64f767636fea9ed035bf85/ext/ffi_c/MethodHandle.c#L107

And it should be released.


larskanis:
LGTM - thank you!

PR #823, opened 2020-09-22
Check JRuby version without accessing Gem

headius:
When running with --disable-gems, JRuby still needs to be able to load FFI from stdlib. Accessing the Gem namespace here requires that gems not be disabled.

This patch uses a simple split and integer comparison to compare the first two digits of the JRuby version with 9.3.

See #763 for the original code.

This patch is necessary for JRuby to fully source the FFI Ruby files from the gem (jruby/jruby#6150). Example failures here are due to the MRI suites running without RubyGems enabled, but they end up needing FFI for internal JRuby functionality.

https://travis-ci.org/github/jruby/jruby/builds/729197020

So close!

### DIFF
--- a/lib/ffi.rb
+++ b/lib/ffi.rb
@@ -8,11 +8,11 @@ if RUBY_ENGINE == 'ruby' || RUBY_ENGINE == 'rbx'
 
   require 'ffi/ffi'
 
-elsif RUBY_ENGINE == 'jruby' && Gem::Version.new(RUBY_ENGINE_VERSION) >= Gem::Version.new("9.3.pre")
+elsif RUBY_ENGINE == 'jruby' && (RUBY_ENGINE_VERSION.split('.').map(&:to_i) <=> [9, 3]) >= 0
   JRuby::Util.load_ext("org.jruby.ext.ffi.FFIService")
   require 'ffi/ffi'
 
-elsif RUBY_ENGINE == 'truffleruby' && Gem::Version.new(RUBY_ENGINE_VERSION) >= Gem::Version.new("20.1.0-dev-a")
+elsif RUBY_ENGINE == 'truffleruby' && (RUBY_ENGINE_VERSION.split('.').map(&:to_i) <=> [20, 1, 0]) >= 0
   require 'truffleruby/ffi_backend'
   require 'ffi/ffi'
 


headius:
The TruffleRuby code might want to do the same.

headius:
The two specific failure cases I see in JRuby CI:

* Our rbconfig/sizeof.rb requires FFI to enumerate native type widths.
```
org/jruby/RubyModule.java:3830:in `const_missing': uninitialized constant Gem (NameError)
	from /home/travis/build/jruby/jruby/lib/ruby/stdlib/ffi.rb:11:in `<main>'
	from org/jruby/RubyKernel.java:1043:in `require'
	from /home/travis/build/jruby/jruby/lib/ruby/stdlib/rbconfig/sizeof.rb:2:in `<main>'
	from org/jruby/RubyKernel.java:1043:in `require'
	from /home/travis/build/jruby/jruby/test/mri/lib/envutil.rb:11:in `<main>'
	from org/jruby/RubyKernel.java:1043:in `require'
	from org/jruby/RubyKernel.java:1071:in `require_relative'
	from /home/travis/build/jruby/jruby/test/mri/lib/test/unit.rb:8:in `<main>'
	from org/jruby/RubyKernel.java:1043:in `require'
	from test/mri/runner.rb:13:in `<main>'
```

* FFI is loaded at boot and used on Windows to implement symlink support and on Solaris to implement flock.
```
NameError: uninitialized constant Gem
  const_missing at org/jruby/RubyModule.java:3830
         <main> at D:/a/jruby/jruby/lib/ruby/stdlib/ffi.rb:11
        require at org/jruby/RubyKernel.java:1043
         <main> at uri:classloader:/jruby/kernel/file.rb:6
           load at org/jruby/RubyKernel.java:1078
         <main> at file:///D:/a/jruby/jruby/lib/jruby.jar!/jruby/kernel.rb:27
```

The latter usage could possibly delay booting FFI, but either way once it's needed it needs to be loadable without RubyGems present.

lib/ffi.rb
```diff
@@ -8,7 +8,7 @@
 
   require 'ffi/ffi'
 
-elsif RUBY_ENGINE == 'jruby' && Gem::Version.new(RUBY_ENGINE_VERSION) >= Gem::Version.new("9.3.pre")
+elsif RUBY_ENGINE == 'jruby' && RUBY_ENGINE_VERSION.split('.').yield_self {|maj, min| maj.to_i >= 9 && min.to_i >= 3}
```
eregon, line 11: An interesting way to compare is `(version.split.map(&:to_i) <=> [9, 3]) >= 0`

eregon:
In TruffleRuby I did not hit this issue because `require "ffi"` loads https://github.com/oracle/truffleruby/blob/4b5eda1a04e669d358bdd2eca6de5abba8fbb8cb/lib/truffle/ffi.rb if there was no `gem "ffi", version` before (or if RubyGems is disabled).
It has the drawback to not load the latest ffi gem with only `require "ffi"` (needs `gem "ffi"` or `bundle exec`).

Making FFI a default gem sounds nice.

lib/ffi.rb
```diff
@@ -8,7 +8,7 @@
...
   JRuby::Util.load_ext("org.jruby.ext.ffi.FFIService")
   require 'ffi/ffi'
```
eregon, line 14: Can you do the same for TruffleRuby with:
```ruby
elsif RUBY_ENGINE == 'truffleruby' && (RUBY_ENGINE_VERSION.split('.').map(&:to_i) <=> [20, 1, 0]) >= 0
```

headius:
Patch has been updated to use a more terse version and remove remaining references to Gem.

eregon (approved):
Looks good to me.
@larskanis Could you merge? (maybe @headius needs a release too?)

headius:
Once there's a release with these changes I can flip the switch to make ffi a default gem in JRuby and delete our copies of the ruby files.

PR #829, opened 2020-09-27
Adds method to check of pointer size if bounded

ankane:
Hi, thanks for this great library!

This PR adds a method to tell if the pointer size is bounded.

```ruby
ptr = FFI::Pointer.new(0)
ptr.bounded?              # false
ptr.slice(0, 10).bounded? # true
```

This can be useful for libraries that accept `FFI::Pointer` arguments.

```ruby
def load_image(ptr, ...)
  raise "Cannot load image from pointer with unbounded size" unless ptr.bounded?
  # do something with size ...
end
```

### DIFF
--- a/ext/ffi_c/Pointer.c
+++ b/ext/ffi_c/Pointer.c
@@ -315,6 +315,22 @@ ptr_address(VALUE self)
     return ULL2NUM((uintptr_t) ptr->memory.address);
 }
 
+/*
+ * Document-method: size_limit?
+ * call-seq: ptr.size_limit?
+ * @return [Boolean]
+ * Return +true+ if +self+ has a size limit.
+ */
+static VALUE
+ptr_size_limit(VALUE self)
+{
+    Pointer* ptr;
+
+    Data_Get_Struct(self, Pointer, ptr);
+
+    return ptr->memory.size == LONG_MAX ? Qfalse : Qtrue;
+}
+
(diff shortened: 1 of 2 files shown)


eregon:
This sounds like a good idea to me.

Right now FFI::Pointer#size returns LONG_MAX for unbounded pointers.
But maybe it should return `nil` instead to not leak that implementation detail?
That could break some things that rely on `#size` always being an Integer though.

ankane:
Thanks @eregon. I think `nil` or `0` would be more intuitive. Currently, code that works well with a `FFI::MemoryPointer` could end up reading too much data if passed an unbounded `FFI::Pointer` (for C functions that take a pointer and bytesize).

larskanis:
Returning `nil` would be more intuitive, but I worry about compatibility. I don't like the name `bounded?` - maybe because I'm not a native English speaker. How about `size_limit?` or `has_size?` .

ankane:
Hey @larskanis, thanks for the response! I thought `bounded?` might be good since `cannot duplicate unbounded memory area` is a [current error message](https://github.com/ffi/ffi/blob/7ff53479e060475672d34ed94714742f343041e7/ext/ffi_c/Pointer.c#L159). Just pushed `size_limit?` and fixed the merge conflict.

ankane:
Thanks @larskanis!

PR #830, opened 2020-10-10
Specify Pointer#order and Struct#order

chrisseaton:
Also slight clarify documentation on `Pointer#order`, raise on an unknown order, and specify `Platform::BYTE_ORDER` etc.

Working towards https://github.com/oracle/truffleruby/issues/2054.

### DIFF
--- a/spec/ffi/struct_spec.rb
+++ b/spec/ffi/struct_spec.rb
@@ -1026,4 +1026,58 @@ describe "variable-length arrays" do
     expect { expect(s[:data][1]).to == 0x12345678 }.to raise_error(IndexError)
   end
 end
+
+describe "Struct order" do
+  before :all do
+    @struct = Class.new(FFI::Struct) do
+      layout :value, :int32
+    end
+  end
+
+  before :each do
+    @pointer = @struct.new
+    @pointer.pointer.write_string("\x1\x2\x3\x4")
+    @pointer
+  end
+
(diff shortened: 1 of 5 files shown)


spec/ffi/struct_spec.rb
```diff
@@ -1026,4 +1026,58 @@ class BuggedStruct < FFI::Struct
...
+describe "Struct order" do
+  before :all do
+    @struct = Class.new(FFI::Struct) do
+      layout :value, :int32
+    end
+  end
+
+  before :each do
+    @pointer = @struct.new
+    @pointer.pointer.write_string("\x1\x2\x3\x4")
```
larskanis, line 1039: I think this should be `write_bytes` since `write_string` [might append](https://github.com/ffi/ffi/pull/806) a final \0 byte in future ffi releases.

larskanis (changes requested):
This was on my agenda, but you solved it better than I would have!

spec/ffi/struct_spec.rb
chrisseaton: Done.

larskanis:
Thank you Chris!

eregon:
This doesn't pass CI on TruffleRuby, due to not being implemented yet: https://github.com/oracle/truffleruby/issues/2054
It seems best to skip those specs on TruffleRuby until implemented to keep the CI green.

PR #834, opened 2020-10-30
Define endian constants and fix CI on TruffleRuby

eregon:
Follow-up of https://github.com/oracle/truffleruby/issues/2054.

@larskanis https://github.com/ffi/ffi/blob/234ebce58eb179035c1c0568f198d34b75f7eb90/ext/ffi_c/rbffi_endian.h seems more verbose/complicated than the Ruby code to define these 3 constants, should we remove that file and always let Ruby code define them? The specs already check that LITTLE_ENDIAN is 1234 and BIG_ENDIAN is 4321, so at least that's valid for all platforms tested in CI.
Also the `pack()` check seems more reliable than a series of `#ifdef`.
We still need `unless defined?` checks for JRuby which defines them in the Java extension.

cc @chrisseaton

### DIFF
--- a/lib/ffi/platform.rb
+++ b/lib/ffi/platform.rb
@@ -144,6 +144,12 @@ module FFI
       "#{LIBPREFIX}c.#{LIBSUFFIX}"
     end
 
+    LITTLE_ENDIAN = 1234 unless defined?(LITTLE_ENDIAN)
+    BIG_ENDIAN = 4321 unless defined?(BIG_ENDIAN)
+    unless defined?(BYTE_ORDER)
+      BYTE_ORDER = [0x12345678].pack("I") == [0x12345678].pack("N") ? BIG_ENDIAN : LITTLE_ENDIAN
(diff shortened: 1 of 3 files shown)


eregon:
Ah, but of course the `.c` files also use those constants, so we need them for the C extension too.
Then the only question is do we keep or remove these 3 lines:
https://github.com/ffi/ffi/blob/234ebce58eb179035c1c0568f198d34b75f7eb90/ext/ffi_c/Platform.c#L71-L73

eregon:
`system-libffi (ubuntu, --disable-system-libffi)` fails but it seems to already happen since a while: https://github.com/ffi/ffi/runs/1323331037?check_suite_focus=true, and unrelated to this PR.

larskanis (commented):
The definition in the header file isn't nice, but pushing the Ruby constants back to C wouldn't be either. So I think your proposal is the best solution for now.

Issue #857, opened 2020-12-19
1.14.0 breaks write_string semantics


skaes:
I find the change that automatically adds a null byte to the target of a write_string call somewhat problematic.

Previously, write_string(s, n) wrote exactly n bytes to the destination, now it writes n+1 bytes. This has the potential to break quite a few programs in horrible ways, especially those who allocate a buffer of s.byte_size before calling write_string(s, s.byte_size) on it. One example can be found here: https://github.com/chuckremes/ffi-rzmq/blob/master/lib/ffi-rzmq/message.rb#L126-L135

I'm wondering whether it would be better to introduce a new method with the new semantics and leave the existing method as is.

eregon:
You can use `write_bytes` to avoid the final null byte.
In any case, this seems worth fixing in `ffi-rzmq`.

If the size of the FFI::Pointer is known, you'll get a nice exception telling you to use `write_bytes` instead.

But https://github.com/chuckremes/ffi-rzmq/blob/740a413ce5a13104fb06702bade68f86fa7626b2/lib/ffi-rzmq/message.rb#L126-L135 calls `malloc()` directly, so FFI doesn't know the size of the returned pointer
(although I guess potentially FFI could treat malloc specially and give the passed argument as size of the pointer).

Some background behind the change in https://github.com/ffi/ffi/issues/805 and https://github.com/ffi/ffi/pull/806.

skaes:
@eregon I already created a PR for ffi-rzmq. That wasn't really why I raised the issue. I guess I had assumed that such a breaking change would be something to introduce with a 2.x release, if at all.

eregon:
The usage linked in `ffi-rzmq` is:
https://github.com/chuckremes/ffi-rzmq/blob/740a413ce5a13104fb06702bade68f86fa7626b2/lib/ffi-rzmq/message.rb#L129
```ruby
      data_buffer.write_string bytes, len
```

I also found two similar usages of `write_string_length` in OptCarrot:
https://github.com/mame/optcarrot/blob/5d7a56e13ae69b4739ae83f28a76aff311baedf7/lib/optcarrot/driver/sdl2_audio.rb#L52-L55
```ruby
        stream.write_string_length(@buff, buff_size)
        @buff.clear
      else
        stream.write_string_length(@buff, stream_len)
```

All 3 usages clearly want to write `len` bytes and nothing more.
[...]

PR #858, opened 2020-12-19
Warn and do not append \0 for FFI::Pointer#write_string(str, str.bytesize)

eregon:
* Fixes https://github.com/ffi/ffi/issues/857


eregon:
I think #859 is better, closing this.

PR #859, opened 2020-12-19
Do not append \0 for FFI::Pointer#write_string(str, str.bytesize)

eregon:
* Fixes https://github.com/ffi/ffi/issues/857

### DIFF
--- a/lib/ffi/pointer.rb
+++ b/lib/ffi/pointer.rb
@@ -101,13 +101,8 @@ module FFI
     # In both cases a final \0 byte is written after the string.
     def write_string(str, len=nil)
       if len
-        if len == size
-          warn "[DEPRECATION] Memory too small to write a final 0-byte in #{caller(1, 1)[0]}. This will raise an error in ffi-2.0. Please use write_bytes instead or enlarge the memory region."
-          write_bytes(str, 0, len)
-        else
-          put_char(len, 0) # Check size before writing str
-          write_bytes(str, 0, len)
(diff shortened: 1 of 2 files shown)


eregon:
@larskanis I think this is ready now, feel free to tweak the warning message as you like.
It seems best to release FFI 1.14.1 soon with this fix to avoid hitting more problematic cases.

skaes:
I have a question regarding the deprecation warnings: if I use a gem which causes the deprecation warning and over which I have no control and the maintainer is busy with other things, can I turn the deprecation warning off?

eregon:
@skaes You can use `$VERBOSE = nil` or `-W0` on the command line to suppress all warnings.
There is also the possibility to filter warnings from `Kernel#warn` by overriding `Warning.warn`, e.g. with https://github.com/jeremyevans/ruby-warning in Ruby 2.5+.
In Ruby 3+ it seems possible to use `warn "message", category: :deprecated`.

skaes:
In my opinion, tying gem specific warnings to the global Ruby warning level is way too coarse grained. And can actually cause real problems in production due to increased log volume. For example, we had to roll back ruby 2.7 upgrades just for that reason: a gem causing a Ruby warning in an inner loop which stressed central log collection and made the log output really hard to use. Of course, one could turn off warnings in production, but usually one wants to leave them in for the application code, but turned off for third party gems.

Anyway: thx for addressing the problem so quickly!

PR #883, opened 2021-02-19
Name the async_cb_thread for easier debugging

ivoanjo:
I found this useful while reviewing datadog/dd-trace-rb#1371 -- we were searching for specs that leaked threads, and saw a "mysterious" thread with no stack that we couldn't exactly point out what created it or why it existed.

By setting a name on the callback thread, it becomes immediately obvious when looking at `Thread.list` what this thread is and where it comes from.

(Thread naming has been around [since Ruby 2.3](https://github.com/ruby/ruby/blob/v2_3_0/NEWS) which means we can do it safely for every supported version)

### DIFF
--- a/spec/ffi/async_callback_spec.rb
+++ b/spec/ffi/async_callback_spec.rb
@@ -39,10 +39,10 @@ describe "async callback" do
     skip "not yet supported on TruffleRuby" if RUBY_ENGINE == "truffleruby"
     skip "not yet supported on JRuby" if RUBY_ENGINE == "jruby"
 
-    thread_name = nil
+    callback_runner_thread = nil
 
-    LibTest.testAsyncCallback(proc { thread_name = Thread.current.name }, 0)
+    LibTest.testAsyncCallback(proc { callback_runner_thread = Thread.current }, 0)
 
-    expect(thread_name).to eq("FFI::Function Callback Runner")
+    expect(callback_runner_thread.name).to eq("FFI::Function Callback Runner")
   end
 end
--- a/spec/ffi/async_callback_spec.rb
+++ b/spec/ffi/async_callback_spec.rb
@@ -34,4 +34,15 @@ describe "async callback" do
     expect(called).to be true
     expect(v).to eq(0x7fffffff)
   end
+
+  it "sets the name of the thread that runs the callback" do
+    skip "not yet supported on TruffleRuby" if RUBY_ENGINE == "truffleruby"
+    skip "not yet supported on JRuby" if RUBY_ENGINE == "jruby"
+
+    thread_name = nil
+
+    LibTest.testAsyncCallback(proc { thread_name = Thread.current.name }, 0)
+
+    expect(thread_name).to eq("FFI::Function Callback Runner")
+  end
 end
--- a/ext/ffi_c/Function.c
+++ b/ext/ffi_c/Function.c
@@ -310,7 +310,7 @@ function_init(VALUE self, VALUE rbFunctionInfo, VALUE rbProc)
 #if defined(DEFER_ASYNC_CALLBACK)
         if (async_cb_thread == Qnil) {
             async_cb_thread = rb_thread_create(async_cb_event, NULL);
-            // Name thread, for better debugging
+            /* Name thread, for better debugging */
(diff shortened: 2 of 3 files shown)


larskanis:
To name the thread is a great idea! It's MRI specific, so that JRuby and TruffleRuby fail. I think it's useful to name the [callback thread](https://github.com/ffi/ffi/blob/3f862ca78ec4e6c2a5eabca8a5f7780d1774c1a0/ext/ffi_c/Function.c#L528) as well. I think as "FFI::Function Callback".

ivoanjo:
Thanks for the feedback!

> It's MRI specific, so that JRuby and TruffleRuby needs to be excluded.

Yeap, already skipped both. It's weird that CI for JRuby seemed to pass even without my change, because locally I did see that it didn't work on JRuby.

> I think it's useful to name the [callback thread](https://github.com/ffi/ffi/blob/3f862ca78ec4e6c2a5eabca8a5f7780d1774c1a0/ext/ffi_c/Function.c#L528) as well. Maybe as "FFI::Function Callback"?

Sounds good! I'll look into it -- I'm guessing it needs a bit of wrapping, because the `Thread` gets started with the callback directly, so I'll probably need to wrap the callback with the equivalent of a proc that first sets the name on the thread it runs on, and then goes on to run the callback.

larskanis:
> > It's MRI specific, so that JRuby and TruffleRuby needs to be excluded.
> 
> Yeap, already skipped both. It's weird that CI for JRuby seemed to pass even without my change, because locally I did see that it didn't work on JRuby.

Unfortunately github actions [doesn't support allow_failure](https://github.com/actions/toolkit/issues/399) so far, so that we use `continue-on-error` as a workaround.
 
> > I think it's useful to name the [callback thread](https://github.com/ffi/ffi/blob/3f862ca78ec4e6c2a5eabca8a5f7780d1774c1a0/ext/ffi_c/Function.c#L528) as well. Maybe as "FFI::Function Callback"?
> 
> Sounds good! I'll look into it -- I'm guessing it needs a bit of wrapping, because the `Thread` gets started with the callback directly, so I'll probably need to wrap the callback with the equivalent of a proc that first sets the name on the thread it runs on, and then goes on to run the callback.

I think it's OK to set the thread name from the thread calling `rb_thread_create()` through the result of this function. The same way as for the dispatcher. I didn't find any information about thread safety of `Thread#name=` but at least the corresponding pthread function is [labled thread save](https://man7.org/linux/man-pages/man3/pthread_setname_np.3.html) and the sample program sets the name from the main thread not from the created thread.

ivoanjo:
> I think it's useful to name the callback thread as well. Maybe as "FFI::Function Callback"?

Done! I ended up using "Callback Runner" since it was a bit weird to have a "FFI::Function Callback" and a "FFI::Function Callback Dispatcher", but let me know if you still prefer your original suggestion :)

ivoanjo:
Hmmm I guess I have some debugging to do on windows. I'm guessing it may be due to the racy manner in which we set the name in one thread and then access it in the other.

spec/ffi/async_callback_spec.rb
```diff
@@ -34,4 +34,15 @@ module LibTest
...
     expect(v).to eq(0x7fffffff)
   end
+
+  it "sets the name of the thread that runs the callback" do
+    skip "not yet supported on TruffleRuby" if RUBY_ENGINE == "truffleruby"
+    skip "not yet supported on JRuby" if RUBY_ENGINE == "jruby"
+
+    thread_name = nil
+
+    LibTest.testAsyncCallback(proc { thread_name = Thread.current.name }, 0)
```
larskanis, line 44: It should be race-free to store `Thread.current` here and call `Thread#name` below. The name is readable even if the thread has terminated.

ivoanjo:
https://github.com/ffi/ffi/blob/master/spec/ffi/fixtures/FunctionTest.c#L128
Hah! Nice troll! 

I can try to name the thread all day long! The "async" callback is not async on windows xD

Took me a bit to debug this ahahaha :)

ivoanjo:
Thanks for picking up the last bit @larskanis! I admit I was not having much sucess poking at the win32 bits and fighting with `_beginthreadex` -- I had gotten so far as to make it work on Ruby 2.7 and 2.6, but was getting segmentation faults on older Rubies.

:bow: thanks for the help!

larskanis:
No problem. Thank you for contributing this!

Issue #732, opened 2019-12-27
FFI segfault when a varargs function is given a callback as a parameter


LikeLakers2:
Hi there! I'm attempting to make a FFI binding for SQLite3, and I ran into a big problem that seems to stem from FFI itself.

Specifically, I seem to be getting a segfault whenever a attached function, with `:varargs` as one of the arguments, receives a callback as one of the parameters.

The code below is the minimal code needed to replicate this. This code, when run as is, currently produces a segfault.
```ruby
require 'ffi'

module SQLite3
	extend FFI::Library
	ffi_lib 'sqlite3'
	
	callback :sqlite3_config_log_xFunc, [:pointer, :int, :string], :void
	attach_function :sqlite3_config, [:int, :varargs], :int
end

def sqlog(data, op, str)
	print 'SQLITE: '
	p [data, op, str]
end

p SQLite3.sqlite3_config(
	16, # SQLITE_CONFIG_LOG
	:sqlite3_config_log_xFunc, method(:sqlog)
)
```

To ensure this wasn't just me being stupid (or SQLite being stupid), I tried changing the parameter list for `sqlite3_config` around a bit.

When the parameter list is `[:int, :sqlite3_config_log_xFunc, :varargs]` (and the function call is edited appropriately), FFI will still segfault -- and in fact, will **always** segfault even when you don't specify the callback parameter in the function call.

However, when the parameter list is `[:int, :sqlite3_config_log_xFunc]` (with no `:varargs`), the function will work as intended. Even if I enter in invalid parameters to `sqlite3_config` (assuming their type matches up with what FFI expects the parameters to be), I no longer get a segfault from FFI, and SQLite receives the function call as expected. Except, now I've lost access to all the other config options that I'm meant to change via `sqlite3_config`, as the parameter list isn't compatible with any of the other configuration verbs offered for `sqlite3_config`.

P.S.  I am running `ffi (1.11.3 x64-mingw32)` on Windows 10, as well as SQLite 3.30.1, in case the version numbers make any difference (they shouldn't, though).

larskanis:
There has been some trouble with callbacks on Windows in the past. Could you try if it works with older ffi versions? And with
```
ridk exec pacman -S mingw-w64-x86_64-libffi
gem inst ffi --platform=ruby -- --enable-system-libffi
```

LikeLakers2:
[...]

PR #885, opened 2021-02-25
Allow to pass callbacks in varargs

vincentisambart:
As I explained in https://github.com/ffi/ffi/issues/732#issuecomment-784703145, if you passed callbacks in varargs, a `NULL` pointer would be dereferenced and your program would crash.
I modified `variadic_invoke` to pass proper `callbackParameters` and `callbackCount` to `rbffi_SetupCallParams`, and that seems to fix the problem.

### DIFF
--- a/ext/ffi_c/Variadic.c
+++ b/ext/ffi_c/Variadic.c
@@ -168,7 +168,8 @@ variadic_invoke(VALUE self, VALUE parameterTypes, VALUE parameterValues)
     ffi_type* ffiReturnType;
     Type** paramTypes;
     VALUE* argv;
-    int paramCount = 0, fixedCount = 0, i;
+    VALUE* callbackParameters = NULL;
+    int paramCount = 0, fixedCount = 0, callbackCount = 0, i;
     ffi_status ffiStatus;
     rbffi_frame_t frame = { 0 };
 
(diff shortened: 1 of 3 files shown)


vincentisambart:
@larskanis Hi! Would you have anything against allowing to pass callbacks via varargs arguments?

vincentisambart:
By the way, passing callbacks in varargs might look like a rare use case, but functions of very used libraries like libcurl or sqlite use it.

For example, libcurl's `curl_easy_getinfo` function is declared as follows:
```c
CURL_EXTERN CURLcode curl_easy_getinfo(CURL *curl, CURLINFO info, ...);
```

A library called [ethon](https://github.com/typhoeus/ethon) was declaring its arguments as `[:pointer, :info, :pointer]`, even though the correct definition should be `[:pointer, :info, :varargs]`.
That was working properly on many platforms, as using varargs or not had a similar ABI, but on ARM Macs, the ABI is different so that does not work properly. A bit more details in typhoeus/ethon#180.

The correct arguments definition needs this PR's fix to work properly.

larskanis:
@vincentisambart Thank you for implementing this!

I merged your PR and did some fine tuning on the master branch. Let me know, if I misunderstood something! I'll try to make a new release soon, but I have only little spare time currently.

vincentisambart:
Thanks a lot!
Sorry for the fix you had to do. It's been too long since last time I wrote C code and completely forgot about freeing memory 😅
I tried the latest master branch with my use case and it still seems to work well 😄

No worries, I totally understand spare time is a limited resource.
Looking forward to the release 😁

jdarnok:
Thanks @vincentisambart and @larskanis for the great job you are doing here! one question regarding a release - can I know how long we have to wait for the release? there is quite a lot of people being blocked by it, including me 😄

Issue #889, opened 2021-03-09
Support for loading versioned symbols


lwoggardner:
Hello

I'm looking to provide ffi bindings for libfuse (to replace c-extension in rfuse)

However libfuse2.x has some funky, possibly broken, symbol versions that are only fixed in libfuse3 which is not necessarily available everywhere yet.

Is there a way to tell FFI to load a versioned symbol?  (dlvsym vs dlsym ?)

For below I'm not sure if FFI is finding 200 (the first entry) or 201 (the unversioned entry) but neither is correct, I need 202 fuse_new@@FUSE_2.6

Assuming it is not possible already would it make sense to patch FFI to allow versioned lookups?

eg something like
```
attach_function :fuse_new, [..args...], :int,  version: '@@FUSE_2.6'
```
If so I'd be happy to have a crack at that.

readelf -Ws /lib/x86_64-linux-gnu/libfuse.so.2.9.9 | grep fuse_new
   200: 0000000000012880    72 FUNC    GLOBAL DEFAULT   15 fuse_new@FUSE_2.5
   201: 0000000000012830    20 FUNC    GLOBAL DEFAULT   15 fuse_new
   202: 0000000000012710    12 FUNC    GLOBAL DEFAULT   15 fuse_new@@FUSE_2.6
   203: 0000000000012820    15 FUNC    GLOBAL DEFAULT   15 fuse_new@FUSE_2.2
   216: 0000000000012850    39 FUNC    GLOBAL DEFAULT   15 fuse_new_compat1@@FUSE_2.2
   218: 0000000000012830    20 FUNC    GLOBAL DEFAULT   15 fuse_new_compat2@@FUSE_2.2
   228: 0000000000012820    15 FUNC    GLOBAL DEFAULT   15 fuse_new_compat22@@FUSE_2.5
   230: 0000000000012880    72 FUNC    GLOBAL DEFAULT   15 fuse_new_compat25@@FUSE_2.6
`

larskanis:
I just searched glibc forth and back to find a way to pass the version through the symbol name, like it's possible for stdcall name decorations on Win32. But there is no such way in glibc. So we need to implement dlvsym into FFI to allow the version selection.

As a workaround for current FFI versions, it's possible to call `libdl` manually like so:
```ruby
module Dl
  extend FFI::Library
  ffi_lib 'libdl'
  attach_function :dlopen, [:string, :int], :pointer
  attach_function :dlvsym, [:pointer, :string, :string], :pointer
end

[...]

PR #890, opened 2021-03-10
add support for looking up versioned functions re #889

lwoggardner:


tduehr (changes requested):
Please add tests.

larskanis:
I wrote some tests in the meantime. They yet need to be excluded on platforms that don't support versioned symbols.

Moreover I tried to build a test that uses similar symbols as shown in https://github.com/ffi/ffi/issues/889 . In case of `fuse_new` there are distinct function entries (12830 vs. 12710) for `fuse_new` and `fuse_new@@FUSE_2.6`  (the default for non versioned `dlsym()` calls). I was unable to mimic this in our `libtest.so`, so far. So the tests don't verify this particular issue.

```
201: 0000000000012830 20 FUNC GLOBAL DEFAULT 15 fuse_new
202: 0000000000012710 12 FUNC GLOBAL DEFAULT 15 fuse_new@@FUSE_2.6
```

lwoggardner:
> Moreover I tried to build a test that uses similar symbols as shown in #889 . In case of `fuse_new` there are distinct function entries (12830 vs. 12710) for `fuse_new` and `fuse_new@@FUSE_2.6` (the default for non versioned `dlsym()` calls). I was unable to mimic this in our `libtest.so`, so far. So the tests don't verify this particular issue.
> 

I'm pretty sure use of versions like that is non standard and they fixed it in Fuse 3.  I think the test are sufficient for the expected use of versioned symbols via dlvsym and I can confirm that it works for the Fuse case as well.

ext/ffi_c/DynamicLibrary.c
```diff
@@ -27,9 +27,14 @@
...
+
+// ruby.h first to provide _GNU_SOURCE which gives us dlvsym if available
+#include <ruby.h>
+
 #include <sys/types.h>
 #include <stdio.h>
 #include <stdint.h>
+
 #if (defined(_WIN32) || defined(__WIN32__)) && !defined(__CYGWIN__)
 # include <winsock2.h>
```
chenqiongqiong, line 39: ,m,,,,

lwoggardner: Not sure what ^^ means but the alternative to this change in include order is to define _GNU_SOURCE specifically and accept that ruby.h generates a warning about it being redefined.

PR #891, opened 2021-03-12
Update README with instructions for submodules

bbuchalter:
This is coming up more as people are experimenting with the HEAD of repos trying to get Apple Silicon support working. Example: https://github.com/grpc/grpc/issues/25669#issuecomment-796563819


larskanis:
Thank you!

PR #895, opened 2021-04-02
Add option for library search paths.

ioquatix:
This is trying to address the same code/TODO as https://github.com/ffi/ffi/pull/616

The difference is, I agree that we should be deferring to the OS.

I've retained the legacy behaviour, but as soon as you specify search paths, you nuke the existing "legacy" search paths.

The reason is, we may want to provide reasonable search paths, or use a `xxx_config` command to load library paths. Then, we would pass that in as arguments to `ffi_lib`.

### DIFF
--- a/lib/ffi/library.rb
+++ b/lib/ffi/library.rb
@@ -87,12 +87,27 @@ module FFI
       raise RuntimeError.new("must only be extended by module") unless mod.kind_of?(Module)
     end
 
+    # The default search paths.
+    DEFAULT_SEARCH_PATHS = ['/usr/lib/','/usr/local/lib/','/opt/local/lib/'].freeze
+
+    # @param name the name of the library to search for
+    # @param search_paths the search paths to look within
+    def find_library_path(libname, search_paths)
+      search_paths.each do |search_path|
+        full_path = File.join(search_path, libname)
+        if File.exist?(full_path)
+          return full_path
+        end
+      end
+
+      return nil
+    end
 
     # @param [Array] names names of libraries to load
     # @return [Array<DynamicLibrary>]
     # @raise {LoadError} if a library cannot be opened
     # Load native libraries.
-    def ffi_lib(*names)
+    def ffi_lib(*names, search_paths: DEFAULT_SEARCH_PATHS)
       raise LoadError.new("library names list must not be empty") if names.empty?
 
       lib_flags = defined?(@ffi_lib_flags) ? @ffi_lib_flags : FFI::DynamicLibrary::RTLD_LAZY | FFI::DynamicLibrary::RTLD_LOCAL
@@ -124,13 +139,9 @@ module FFI
               if ldscript
                 retry
               else
-                # TODO better library lookup logic
                 unless libname.start_with?("/") || FFI::Platform.windows?
-                  path = ['/usr/lib/','/usr/local/lib/','/opt/local/lib/'].find do |pth|
-                    File.exist?(pth + libname)
-                  end
-                  if path
-                    libname = path + libname
+                  if path = find_library_path(libname, search_paths)
+                    libname = path
                     retry
                   end
                 end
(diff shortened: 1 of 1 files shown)


ioquatix:
I would also like to consider a potential extension like this:

```
load_library 'mariadb', config: 'mysql_config'
```

This would try to load `mariadb` using `dlopen`, and if that fails, the. executes `mysql_config --libs` and extracts `-Lpath` and `-l` library.

Do you think you'd be willing to accept something like this?

ioquatix:
I've added an experimental commit which lets you load from config tools:

```
irb(main):001:0' require 'ffi'
=> true
irb(main):002:1* module M
irb(main):003:1*   extend FFI::Library
irb(main):004:0> end
=> M
irb(main):005:0> M.load_library_from_config('mysql_config')
=> [#<FFI::DynamicLibrary:0x00007ffe0588b368 @name="/opt/local/lib/mariadb-10.2/mysql/libmariadb.dylib">]
```

I think this is potentially a really great solution for situations where you need to dynamically determine the library.

The way you'd use it is something like this:

```
module M
  extend FFI::Library
  
  load_library('mariadb') or load_library_from_config('mysql_config')
end
```

ioquatix:
I'd like to see us move away from using `defined?` if possible as it's an extra dimension to go wrong, so I've added `@ffi_libs` instance variable setup as part of extending the module.

ioquatix:
I played around with the interface a bit further and made a few observations:

- I think I'd prefer if all methods start with `ffi` as in `ffi_attach_function` etc.
- `ffi_lib` is overloaded with functionality, and I think I want it to be simpler.
- There are different ways to load libraries using things like `pkg-config` or `xyz_config`. These should be supported but I don't think there is a "default" approach.
- We don't want to break existing usage of `ffi` so it's hard to make big changes to `FFI::Library`.

In order to explore this a bit further, I've decided to make a gem `ffi-module` which we can work on independently of this gem, and that avoids breaking existing usage of `FFI::Library`. I want to look at introducing extensions e.g. `FFI::Module::ConfigTool` which expose the specific functionality. The base `FFI::Module` is very similar to `FFI::Library` but it (IMHO) has a much cleaner and consistent interface.

Going forward, I don't mind how we work on this, whether we can merge this work some how, or whether it makes sense to keep it separate. I certainly think there is value in "cleaning up" `FFI::Library` but maybe it's impossible, and honestly, it does work and it's okay, it's just that every non-trivial `ffi` integration I've seen has pretty complex integrations with `ffi_lib` and I think this is something I'd like to see simplified.

ioquatix:
Okay, here is the repo: https://github.com/ioquatix/ffi-module

I've split up the logic into separate modules:

- `FFI::Module::Library` - the core interface for libraries to attach functions, variables, define types etc.
- `FFI::Module::Loader` - expose a simple interface for loading libraries directly.
- `FFI::Module::ConfigTool` - expose an interface for loading libraries from tools like `pkg-config` or `xyz_config`.

I found one bug in `ffi`:

```
samuel@Fukurou ~/D/i/ffi-module (master)> irb
irb(main):001:0> module FFI; end; p FFI.object_id
260
=> 260
irb(main):002:0> require 'ffi'; p FFI.object_id
300
=> 300
```

It won't reuse an existing `module FFI` once you `require 'ffi'`.

ioquatix:
Okay, so after playing around with this, I ended up with this:

https://github.com/socketry/db-mariadb/blob/f805b9a2dade680e5926081a4b009655a78b82b0/lib/db/mariadb/native.rb#L21-L34

and this:

https://github.com/socketry/db-postgres/blob/2b21928acd0168d8884f56000ac18f36bee05356/lib/db/postgres/native.rb#L21-L31

I am thinking to add one more final statement, `ffi_load_failure(message)` which fails with LoadError and logs the given message, which should explain to the user how to make it work (i.e. install the following packages, etc).

ioquatix:
Here is a picture of the load error that I put together:

image

eregon:
One potential issue with `FFI::Module` is it can be confusing with the top-level `Module` class.
This issue can affect any code under `module FFI`.
From experience with similar issues in TruffleRuby, it can be very confusing (the conclusion is never define a top-level constant as nested too).

PR #897, opened 2021-04-15
Allow overriding `gcc` with the `CC` env var in `const_generator.rb` and `struct_generator.rb`

leshik:
Right now one can override `gcc` for `types_generator.rb`, but not `const_generator.rb` and `struct_generator.rb`. This PR fixes this.

Related: https://github.com/typhoeus/ethon/issues/186

### DIFF
--- a/lib/ffi/tools/const_generator.rb
+++ b/lib/ffi/tools/const_generator.rb
@@ -124,7 +124,8 @@ module FFI
         f.puts "\n\treturn 0;\n}"
         f.flush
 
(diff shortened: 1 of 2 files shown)


larskanis:
There could be more compiler independence in this part of the code, probably.

It also makes the compiler a runtime dependency of the gem, which most ffi based gems don't have otherwise. That's why these classes are rarely used. On the other hand ffi.gem needs a compiler anyway in most cases.

leshik:
> On the other hand ffi.gem needs a compiler anyway in most cases.

@larskanis Right, but it's usually a good idea to separate build & runtime. E.g. in my case, `gcc` is used on `bundle install` stage to build things like `nokogiri` and `ffi`, but it's missing at runtime. Now I added `tcc` to runtime as it adds only 100kb overhead to the Docker image.

Baltazardoung:
Can you tell us when a new version (`v1.15.1` ?) including this patch will be released ? Thanks a lot.

PR #912, opened 2021-08-12
Correct module lookup when including ffi-module

danielevans:
The inclusion of @ioquatix's `ffi-module` gem to a project causes errors loading other ffi gems (such as `rbnacl`) due to the definition of `FFI::Module` which makes `.kind_of?(Module)` equivalent to `.kind_of?(FFI::Module)`.

This change simply changes that lookup to ensure that it is checking against the top-level Module constant.

### DIFF
--- a/lib/ffi/library.rb
+++ b/lib/ffi/library.rb
@@ -84,7 +84,7 @@ module FFI
     # @raise {RuntimeError} if +mod+ is not a Module
     # Test if extended object is a Module. If not, raise RuntimeError.
     def self.extended(mod)
-      raise RuntimeError.new("must only be extended by module") unless mod.kind_of?(Module)
+      raise RuntimeError.new("must only be extended by module") unless mod.kind_of?(::Module)
     end
 
 


ioquatix:
Hmm, I wonder if I should rename the gem. I didn't realise this issue at the time.

eregon:
The change is fine (make the check a bit more robust) but IMHO the much bigger issue is to define a class or module named `Module` which isn't the `::Module` of Ruby core.
So +1000 to rename `FFI::Module` to something else.

ioquatix:
```
FFI::SomethingElse
```

But seriously, I agree, this was kind of an oversight on my part.

That being said, I've encountered this issue in other places.

Sometimes Ruby constant lookup causes pain if someone introduces a new module into an existing one.

larskanis:
Do we need a new release due to this issue?

danielevans:
There is no rush to release the fix from my perspective, the issue is easily worked around and this gem is not widely depended upon.

PR #929, opened 2022-01-03
fix unlink error 

maierru:
when using shared tmp dir inside parallel docker containers (same Process id)
you can see errors like below

```ruby
const_generator.rb:136:in ``': Text file busy - /tmp/rb_const_gen_bin_1 (Errno::ETXTBSY)
```

```ruby
const_generator.rb:137:in `unlink': No such file or directory @ apply2files - /tmp/rb_const_gen_bin_1 (Errno::ENOENT)
```

### DIFF
--- a/lib/ffi/tools/const_generator.rb
+++ b/lib/ffi/tools/const_generator.rb
@@ -105,9 +105,10 @@ module FFI
     # @return [nil]
     # @raise if a constant is missing and +:required+ was set to +true+ (see {#initialize})
     def calculate(options = {})
-      binary = File.join Dir.tmpdir, "rb_const_gen_bin_#{Process.pid}"
+      binary_path = nil
 
       Tempfile.open("#{@prefix}.const_generator") do |f|
+        binary_path = f.path + ".bin"
         @includes.each do |inc|
           f.puts "#include <#{inc}>"
         end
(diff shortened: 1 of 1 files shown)


larskanis:
Can't we use Tempfile for this job like so (or similar):

```ruby
Tempfile.open("#{@prefix}.rb_const_gen_bin") do |bf|
  binary = bf.path
  [...]
```

maierru:
> Can't we use Tempfile for this job like so (or similar):
> 
> ```ruby
> Tempfile.open("#{@prefix}.rb_const_gen_bin") do |bf|
>   binary = bf.path
>   [...]
> ```

thanks for quick response!

As I understood, we need two files - one for generated c source code and another for binary (which is created by gcc)
first one used as you suggested 
https://github.com/ffi/ffi/pull/929/files#diff-4756902fc04469bc839933537bb1bda3bbc1260617e86e215a587bfba0873bcdR128
but main assumption here, that no one is going to create same binary at the end of compilation (because of different process id), but inside docker you start with the process pid 1 and if you share tmp directory that leads to errors :(

maierru:
> Can't we use Tempfile for this job like so (or similar):
> 
> ```ruby
> Tempfile.open("#{@prefix}.rb_const_gen_bin") do |bf|
>   binary = bf.path
>   [...]
> ```

maybe I misunderstood you, I did it in a different way, wdyt?

larskanis:
This looks good. Thank you!

Issue #880, opened 2021-02-05
Could not open library 'sodium': dlopen(sodium, 5): image not found. (LoadError) on Apple Silicon M1


OleMchls:
Hey there,

I'm trying to get `rbnacl` to function on my M1, but I am hitting the following error message:

```
ole@Oles-MBP-2 test % ruby -e "require 'rbnacl'" 
/Users/ole/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/rbnacl-7.1.1/lib/rbnacl/sodium/version.rb:10: warning: already initialized constant RbNaCl::Sodium::Version::MINIMUM_LIBSODIUM_VERSION
/Users/ole/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/rbnacl-7.1.1/lib/rbnacl/sodium/version.rb:10: warning: previous definition of MINIMUM_LIBSODIUM_VERSION was here
/Users/ole/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/rbnacl-7.1.1/lib/rbnacl/sodium/version.rb:11: warning: already initialized constant RbNaCl::Sodium::Version::MINIMUM_LIBSODIUM_VERSION_FOR_ARGON2
/Users/ole/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/rbnacl-7.1.1/lib/rbnacl/sodium/version.rb:11: warning: previous definition of MINIMUM_LIBSODIUM_VERSION_FOR_ARGON2 was here
/Users/ole/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/rbnacl-7.1.1/lib/rbnacl/sodium/version.rb:12: warning: already initialized constant RbNaCl::Sodium::Version::MINIMUM_LIBSODIUM_VERSION_FOR_ARGON2ID
/Users/ole/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/rbnacl-7.1.1/lib/rbnacl/sodium/version.rb:12: warning: previous definition of MINIMUM_LIBSODIUM_VERSION_FOR_ARGON2ID was here
Traceback (most recent call last):
	2: from -e:1:in `<main>'
	1: from /Users/ole/.asdf/installs/ruby/2.7.2/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require'
[...]

PR #936, opened 2022-01-23
Move /opt before /usr to search for M1 native libs first - solve the conflicts when both x64 and arm64 libs installed

southwolf:
This is trying to solve 
https://github.com/ffi/ffi/issues/880#issuecomment-1019394749

In apple silicon machines, when both versions are installed, the native arm64 libs (`'/opt/local/lib/'`) should be the first option, and x64 (`'/usr/local/lib/'`) as a failsafe fallback only when arm64 version is not available.

In my case: I've installed both x64 & arm64 version of glib, but `ffi` found x64 and then stopped, giving up to try `/opt` ones
image

While placing `/opt` before `/usr/local`

image

### DIFF
--- a/lib/ffi/library.rb
+++ b/lib/ffi/library.rb
@@ -126,7 +126,7 @@ module FFI
               else
                 # TODO better library lookup logic
                 unless libname.start_with?("/") || FFI::Platform.windows?
-                  path = ['/usr/lib/','/usr/local/lib/','/opt/local/lib/', '/opt/homebrew/lib/'].find do |pth|
+                  path = ['/usr/lib/','/opt/local/lib/', '/opt/homebrew/lib/', '/usr/local/lib/'].find do |pth|
                     File.exist?(pth + libname)
                   end
                   if path


notjulian:
Hi @southwolf,
with this, or https://github.com/ffi/ffi/pull/941, it is possible to use this lib without rosetta, natively?

thanks 

PS: I use it with react native

southwolf:
Hi @notjulian yes that is the purpose of this PR. If you have installed native ARM64 libraries, it will use native libs first

ronaldtse:
Thank you @southwolf for this!

Can a project maintainer look at this? (ping @larskanis) It is a necessary step for using the ffi gem on Apple M1 natively -- which applies to nearly the full Mac line up today. Thanks!

chinyakao:
Thank you @southwolf !

Could project maintainer (@larskanis) updated this ASAP? 
It is very necessary for MacOs M1 user.
Thank you so much!

southwolf:
@larskanis @Kerilk Hi ffi maintainers could you please take some time to review and merge this PR as a temporary solution if no better one available?

 More and more people are asking about the Apple Silicon compatibility issue. Thanks!

Kerilk:
@rainyjonne, @southwolf I am not a maintainer and shared library are not my subject of predilection, but I'll just throw my two cents here:

The change you propose is impacting most platforms and can break the existing workflow of many. If those are indeed required for those specific platform, I would add a conditional for it, using this directory order only on those specific M1 Macs. You may need to add additional detection mechanisms to: https://github.com/ffi/ffi/blob/master/lib/ffi/platform.rb

southwolf:
@Kerilk Thanks for pointing out! Agree with your point, but I didn't do that because in most cases people just use /usr/local as default installation path, /opt is only seen in MacOS + Apple Silicon scenarios AFAIK. Another reason I made a quick fix instead of a thorough solution is that maintainers of the library just ignored this PR (and all other related PRs/issues) without responses, so maybe I'll just leave it there as a temporary workaround or FYI.

PR #949, opened 2022-04-13
Bitmask improvement

Kerilk:
Hello,

I encountered an issue with the current Bitmask implementation (I wrote those, so I am to blame), when the native type is signed, and a flag uses the most significant bit. The fact that the native value can be negative was causing some incorrect remainders when decoding the flags, and range errors when synthesizing the value from the flags.

This PR fixes the above issue, and adds relevant tests.
One of those tests fails on TruffleRuby: testing UINT64 bitmasks with most significant bit set:
`unsupported type [9223372036854775811]` which corresponds to `0x8000000000000003 == (1<<63) | (1<<1) | (1<<0)`
I assume this is a limitation of TruffleRuby, but wanted to check before adding this to the expected failure list.

I've also taken the opportunity to clean the code up a bit.

Thanks,
Brice

### DIFF
--- a/lib/ffi/enum.rb
+++ b/lib/ffi/enum.rb
@@ -220,7 +220,7 @@ module FFI
     #  @param [Symbol] query
     #  @return [Integer]
     # @overload [](query)
-    #  Get bitmaks value from symbol array
+    #  Get bitmask value from symbol array
     #  @param [Array<Symbol>] query
     #  @return [Integer]
     # @overload [](*query)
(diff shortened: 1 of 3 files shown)


eregon:
Regarding the failure on TruffleRuby, could you simply skip that test on TruffleRuby for the moment?
It seems I need to handle types which `include DataConverter` like `Enum`/`Bitmask` by using their `native_type`, similar to how `Type::Mapped` is handled currently, in TruffleRuby's backend.
The test fails because that value if a Bignum-range Integer, and Truffle NFI can't understand that.

Kerilk:
@eregon, Thanks for the quick feedback. I was unsure of the diagnostic as the signed case did not cause the same trouble, despite also being out of bound for a FixNum. I'll iterate on the fix if other related issues pop.

larskanis:
LGTM, merged! Thank you very much!

PR #955, opened 2022-05-04
Normalize pointer typedef names

postmodern:
Building on PR #953, I also removed the leading `*` character from all typedef names. This normalizes the typedef names for pointer types, such as `caddr_t`. So instead of having `:"*caddr"` on some platforms, we only have `:caddr` across all platforms.


larskanis:
Your patches make sense, but the files are all generated by [this script](https://github.com/ffi/ffi/blob/965c15f0a8219142efa5447e5a97d25049418d80/lib/ffi/tools/types_generator.rb). Can you change it, so that the removed lines are also removed in the generated output?

postmodern:
Hmm, should I omit types that contain `*` or simply default them to a `:pointer` type alias?

Issue #537, opened 2016-11-08
How does one read a value of :size_t from a MemoryPointer?


nathanaeljones:
If I attempt to call `.read(:size_t)`, 
I get this error: `NoMethodError: undefined method `read' for #<FFI::MemoryPointer address=0x007f941a67f090 size=8>`

The only example I can find is here:
https://github.com/ffi/ffi/blob/c52ceb173288073a321a94b401243ff40110d910/spec/ffi/rbx/memory_pointer_spec.rb#L95

PR #958, opened 2022-05-26
Add aliases in AbstractMemory for size_t and ssize_t.

Kerilk:
Hello,

This is a feature I end up re-implementing in most of my gems, so I thought maybe it warranted to be here.
This implements `read_size_t` and `read_ssize_t` and their counterparts in AbstractMemory.

The proposal is rather straightforward, but some aspects may warrant discussion:
 * I kept the `_t` suffix as I didn't want any confusion between `get_size` and `size` methods,
 * I do this at runtime, but as `sizeof(size_t)` is a compile time constant it shouldn't matter,
 * This only support size of 32 and 64 bits, but I think I noticed this assumption in several places in the code,
 * I refactored the ALIAS macro, I avoided using C99 argument macro with no token for maximum portability, but it may feel clunky.

This would also close https://github.com/ffi/ffi/issues/537

Thanks

### DIFF
--- a/ext/ffi_c/AbstractMemory.c
+++ b/ext/ffi_c/AbstractMemory.c
@@ -771,7 +771,7 @@ rbffi_AbstractMemory_Init(VALUE moduleFFI)
      * * read_array_of_uint<i>size</i>(length)
      * where _size_ is 8, 16, 32 or 64. Same methods exist for long type.
      *
-     * Aliases exist : _char_ for _int8_, _short_ for _int16_, _int_ for _int32_ and <i>long_long</i> for _int64_.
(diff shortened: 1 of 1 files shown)


eregon:
Just passing by, this would need to be tested.

Kerilk:
@eregon I though so as well, but I couldn't find where the other aliases are tested. Maybe they are and I just missed it.

eregon:
`struct_spec.rb` has related tests

Kerilk:
@eregon I ended up adding tests to `pointer_spec.rb` as `write_array_of_type` and `read_array_of_type` were somewhat tested there. I hope it is fine.
Truffle ruby seems to be getting those methods from https://github.com/oracle/truffleruby/blob/master/src/main/ruby/truffleruby/core/truffle/ffi/pointer_access.rb so these would have to be added there as well.

eregon:
I think it'd be good to add something like `ptr.read_array(type, length)` (and `write_array`).
The existing `read_array_of_type` feels clunky due to specifying both the type and the method to read/write (redundant arguments).

PR #962, opened 2022-06-17
Fix Pointer#initialize using NUM2LL instead of NUM2ULL

fcheung:
If the high bit of the address was set this would raise RangeError (bignum too big to convert into long long). This is not uncommon on platforms that use the high bits of pointers for purposes such as pointer authentication (eg arm64 on apple platforms - https://developer.apple.com/documentation/security/preparing_your_app_to_work_with_pointer_authentication )

I've not encountered this error directily but I have had reports of it from users with apple M1 hardware

This also now matches Pointer#address which uses ULL2NUM.

### DIFF
--- a/ext/ffi_c/Pointer.c
+++ b/ext/ffi_c/Pointer.c
@@ -112,7 +112,7 @@ ptr_initialize(int argc, VALUE* argv, VALUE self)
     switch (TYPE(rbAddress)) {
         case T_FIXNUM:
         case T_BIGNUM:
-            p->memory.address = (void*) (uintptr_t) NUM2LL(rbAddress);
+            p->memory.address = (void*) (uintptr_t) NUM2ULL(rbAddress);
             p->memory.size = LONG_MAX;
             if (p->memory.address == NULL) {
                 p->memory.flags = 0;
--- a/spec/ffi/pointer_spec.rb
+++ b/spec/ffi/pointer_spec.rb
@@ -244,7 +244,7 @@ describe "Pointer" do
       pointer = FFI::Pointer.new(:uint8, max_address)
       expect(pointer.address).to eq(max_address)
     end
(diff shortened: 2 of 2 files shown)


larskanis:
We probably have to exclude the test on TruffleRuby. @eregon Am I right?

eregon:
Yeah, it looks like it since it doesn't work yet: https://github.com/ffi/ffi/runs/6942461248?check_suite_focus=true
@fcheung Could you add `if RUBY_ENGINE != "truffleruby"` after the second `end` of the new spec?

fcheung:
> Yeah, it looks like it since it doesn't work yet: https://github.com/ffi/ffi/runs/6942461248?check_suite_focus=true
> @fcheung Could you add `if RUBY_ENGINE != "truffleruby"` after the second `end` of the new spec?

Cool - I have done that.

fcheung:
Is there anything more I can do on this?

eregon (approved):
Looks good to me.
The JRuby windows failure is unrelated, ignore it for now.

@larskanis Could you merge?
(and in general are you OK if I merge PRs to FFI I reviewed?)

eregon:
Ah, but it also fails on JRuby despite the status being green:
https://github.com/ffi/ffi/runs/7097427538?check_suite_focus=true

So yes exclude it on JRuby too then. And cc @headius JRuby's ffi has the same issue.

larskanis:
The jruby CI run is marked as continue-on-error. Maybe we should remove this special handling - it is often confusing. Apart from this, this PR is Ok to merge.

fcheung:
I’ve added an exclusion for jruby.

eregon:
@larskanis FYI, I removed the `continue-on-failure` in https://github.com/ffi/ffi/commit/1d448cfa3be018e385ac39bce8cc32dcc27c91e2

PR #965, opened 2022-06-26
prioritize /opt/homebrew/lib (for macos on aarch64 only) for searching path

dzprofessionerd:


ronaldtse:
Since #941 is closed, I hope this can be merged soon.

lib/ffi/library.rb
```diff
@@ -126,7 +126,12 @@ def ffi_lib(*names)
               else
                 # TODO better library lookup logic
                 unless libname.start_with?("/") || FFI::Platform.windows?
-                  path = ['/usr/lib/','/usr/local/lib/','/opt/local/lib/', '/opt/homebrew/lib/'].find do |pth|
+                  path_candidates = if FFI::Platform::ARCH == 'aarch64' && FFI::Platform.mac?
+                    ['/opt/homebrew/lib/','/usr/lib/','/usr/local/lib/','/opt/local/lib/']
+                  else
+                    ['/usr/lib/','/usr/local/lib/','/opt/local/lib/','/opt/homebrew/lib/']
```
eregon, line 132: Could you remove `/opt/homebrew/lib/` from this list? That seems only useful on darwin-arm64.

Also could you add a space after every comma, or even better use `%w[path1 path2]`?

eregon:
In general it's still unclear if we should prefer /opt/homebrew/lib to /usr/lib (probably not) or /usr/local/lib (maybe, but shouldn't be necessary, and no if that's non-homebrew stuff).
I'm thinking the only safe solution here is to use the existing path, and try every part of it, not just the first directory.

eregon:
I made a PR for the general fix: https://github.com/ffi/ffi/pull/968

PR #975, opened 2022-10-14
Tapioca DSL Compiler

jscheid:
This adds a [Tapioca](https://github.com/Shopify/tapioca) DSL compiler for FFI declarations, to generate [RBI files](https://sorbet.org/docs/rbi) with type information for Sorbet.

The first commit exposes some information that the compiler needs to be able to generate types. Without this information the types would be a lot less precise. The second commit adds the compiler itself.

The compiler is picked up automatically by Tapioca when this gem is installed in the project in which Tapioca is run.

I had originally planned to open a PR for only the first commit here and contribute the compiler to the Tapioca repo, but it occurred to me that here is a better place for it. One advantage of having it here is that it prevents people from using the compiler with an older FFI version, which would generate disappointing types.

I have a full test suite for the compiler sitting here but it was written for the Tapioca test harness, I'll have to adapt it to rspec. Before I do that, is there any interest in this at all? And if not, would you accept a PR with just the first commit?

### DIFF
--- a/ext/ffi_c/Function.c
+++ b/ext/ffi_c/Function.c
@@ -376,9 +376,11 @@ static VALUE
 function_attach(VALUE self, VALUE module, VALUE name)
 {
     Function* fn;
+    FunctionType* info;
     char var[1024];
 
     Data_Get_Struct(self, Function, fn);
+    Data_Get_Struct(fn->rbFunctionInfo, FunctionType, info);
 
     if (fn->info->parameterCount == -1) {
         rb_raise(rb_eRuntimeError, "cannot attach variadic functions");
@@ -403,9 +405,12 @@ function_attach(VALUE self, VALUE module, VALUE name)
     rb_define_singleton_method(module, StringValueCStr(name),
             rbffi_MethodHandle_CodeAddress(fn->methodHandle), -1);
 
-
     rb_define_method(module, StringValueCStr(name),
             rbffi_MethodHandle_CodeAddress(fn->methodHandle), -1);
+
+    rb_hash_aset(rb_cv_get(module, "@@ffi_functions"),
+            ID2SYM(rb_intern(StringValueCStr(name))),
+            rb_ary_new3(2, info->rbParameterTypes, info->rbReturnType));
 
     return self;
 }
@@ -458,16 +463,6 @@ function_release(VALUE self)
     fn->closure = NULL;
 
     return self;
-}
-
-static VALUE
-function_type(VALUE self)
-{
-    Function* fn;
-
-    Data_Get_Struct(self, Function, fn);
-
-    return fn->rbFunctionInfo;
 }
 
 static void
@@ -900,7 +895,6 @@ rbffi_Function_Init(VALUE moduleFFI)
     rb_define_method(rbffi_FunctionClass, "call", function_call, -1);
     rb_define_method(rbffi_FunctionClass, "attach", function_attach, 2);
     rb_define_method(rbffi_FunctionClass, "free", function_release, 0);
-    rb_define_method(rbffi_FunctionClass, "type", function_type, 0);
     rb_define_method(rbffi_FunctionClass, "autorelease=", function_set_autorelease, 1);
     /*
      * call-seq: autorelease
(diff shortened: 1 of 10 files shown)


jscheid:
On second thought, and considering that Tapioca has a disclaimer about their API stability and no mechanism for API versioning as far as I can tell, perhaps it's better for the compiler to live in a separate gem for now. It could declare a minimum version on ffi to ensure it can generate good types, and it can declare a dependency on Tapioca to ensure that API versions match.

I'll leave the second commit up for now as an illustration but I suggest the final PR would consist of just the first commit (afaaa53) unless anyone feels strongly another way.

ext/ffi_c/Function.c
```diff
@@ -890,6 +900,7 @@ rbffi_Function_Init(VALUE moduleFFI)
     rb_define_method(rbffi_FunctionClass, "call", function_call, -1);
     rb_define_method(rbffi_FunctionClass, "attach", function_attach, 2);
     rb_define_method(rbffi_FunctionClass, "free", function_release, 0);
+    rb_define_method(rbffi_FunctionClass, "type", function_type, 0);
```
larskanis, line 903: `FFI::FunctionInfo` is an internal class only. It is used for internal optimization, but shouldn't be exposed anywhere. I think `FFI::Function` should get `return_type` and `param_types` instead.

lib/tapioca/dsl/compilers/ffi.rb
```diff
@@ -0,0 +1,196 @@
...
+      #
+      #   func, var, var=
+      #
+      class FFI < Compiler
+        def decorate
+          root.create_path(constant) do |mod|
+            constant.class_variables.each do |class_var|
+              candidate = constant.class_variable_get(class_var)
+              if candidate.is_a?(::FFI::Function)
+                method_name = class_var.to_s.gsub(/\A@*/, "")
```
larskanis, line 35: Accessing internal variables of FFI is risky. I think this is a good reason to add this compiler directly into ffi.

On the other hand I wonder, if we shouldn't make a public API to query this information and for other two cases below. Then I would prefer to store the compiler in another gem.

larskanis (changes requested):

lib/tapioca/dsl/compilers/ffi.rb
eregon: It feels pretty hacky to me, I think new method(s) to query what is needed is much better.
For instance this logic is unlikely to work on the TruffleRuby backend of ffi.

jscheid: Thanks for your feedback both. All good points. I've pushed a change that exposes attached functions, getters and setters along with metadata in the following format (up for discussion):
- For getters and setters, a map from symbol (`:foo` or `:foo=`) to type class (e.g. `FFI::Type::STRING`)
- For functions, a map from symbol to an array of the shape `[param_types, result_type]` mimicking the arguments to `attach_function`, again using type classes.
- For variadic functions, the same with symbol `:varargs` tacked on the end of the `param_types`.

I'm a bit unhappy with `param_types` containing a mix of classes (`FFI::Type::String`) and the `:varargs` symbol. It makes me wonder if maybe the values should contain symbols instead (e.g. `:string`). However, we'd still have a mix when e.g. structure references are used so I'm not sure that would be an improvement. It would also force me to duplicate some of the type resolution code from FFI in the compiler.

Let me know what you think.

jscheid: PS I haven't tested the latest changes to the DSL compiler yet, I've added them only for illustration.

ext/ffi_c/Function.c
jscheid: With the latest changes (see below) this isn't necessary anymore, I've removed it.

lib/tapioca/dsl/compilers/ffi.rb
jscheid: It just occurred to me that since getters and setters _are_ functions, we might as well roll them into the `attached_functions` data, wdyt?

Issue #981, opened 2022-12-11
`Library ffi_lib interprets INPUT() in loader scripts` spec fails on macOS


eregon:
See https://github.com/ffi/ffi/actions/runs/3668712621/jobs/6202029490
```
Failures:

  1) Library ffi_lib interprets INPUT() in loader scripts
     Failure/Error: raise LoadError, [*errors, SEARCH_PATH_MESSAGE].join(".\n")

     LoadError:
       Could not open library '/Users/runner/work/ffi/ffi/spec/ffi/fixtures/ldscript.so': dlopen(/Users/runner/work/ffi/ffi/spec/ffi/fixtures/ldscript.so, 0x0005): tried: '/Users/runner/work/ffi/ffi/spec/ffi/fixtures/ldscript.so' (not a mach-o file).
       Searched in <system library path>, /usr/lib, /usr/local/lib, /opt/local/lib
     # ./lib/ffi/dynamic_library.rb:65:in `load_library'
     # ./lib/ffi/library.rb:103:in `block in ffi_lib'
     # ./lib/ffi/library.rb:102:in `map'
     # ./lib/ffi/library.rb:102:in `ffi_lib'
     # ./spec/ffi/library_spec.rb:99:in `block (4 levels) in <top (required)>'
     # ./spec/ffi/library_spec.rb:97:in `initialize'
     # ./spec/ffi/library_spec.rb:97:in `new'
     # ./spec/ffi/library_spec.rb:97:in `block (3 levels) in <top (required)>'
```

I'm not sure how to fix it.
* Are loader scripts something supported on macOS or is that a Linux-only or not-macos-windows thing?
* The extension of `ldscript.so` should rather be `ldscript.dylib` on macOS, no?
* The regexp to match this is:
```ruby
if ex.message =~ /(([^ \t()])+\.so([^ \t:()])*):([ \t])*(invalid ELF header|file too short|invalid file format)/
```
(unchanged by https://github.com/ffi/ffi/pull/968/files)
so that makes it sound like only `.so` support linker scripts?

It's a mystery to me why this spec passed before #968 on macOS, it sounds like it should never have worked on macOS.

I guess there are two main ways to fix it:
* Don't run this test on macOS if linker scripts are not a thing there anyway
* Try adapting the regexp to accept this `not a mach-o file` error

cc @larskanis

larskanis:
[...]

PR #982, opened 2022-12-12
Add error message of newer Macos to linker script emulation

larskanis:
.. and adjust the regex to match the new format of the message text.
The new error message is raised in macOS-12.6.1 while macOS-11.7 worked.

Fixes https://github.com/ffi/ffi/issues/981


eregon:
https://github.com/ffi/ffi/pull/983 was merged

PR #992, opened 2023-02-17
Store functions in a hash rather than class variables

tenderlove:
Setting class variables will mutate the shape of the receiver, causing more inline cache polymorphism than necessary.  Storing functions in a hash would better serve the intended functionality, and also benefit runtime performance

### DIFF
--- a/ext/ffi_c/Function.c
+++ b/ext/ffi_c/Function.c
@@ -395,10 +395,15 @@ function_attach(VALUE self, VALUE module, VALUE name)
     }
 
     /*
-     * Stash the Function in a module variable so it does not get garbage collected
+     * Stash the Function in a hash so it does not get garbage collected
      */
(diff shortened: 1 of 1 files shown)


ext/ffi_c/Function.c
```diff
@@ -395,10 +395,15 @@ function_attach(VALUE self, VALUE module, VALUE name)
     }
 
     /*
-     * Stash the Function in a module variable so it does not get garbage collected
+     * Stash the Function in a hash so it does not get garbage collected
      */
-    snprintf(var, sizeof(var), "@@%s", StringValueCStr(name));
```
larskanis, line 400: The `var` variable can be removed.

larskanis (changes requested):
Hi Aaron! Tests fail with `uninitialized constant LibTest::FFI_FUNCTIONS`. Could you please fix this?

PR #996, opened 2023-03-02
Convert FFI::ThreadData to TypedData

casperisfine:

The old untyped DATA API is soft deprecated and
this new one open the door to write barriers, compaction, memsize etc.


casperisfine:
Thanks Lars! ❤️

PR #1000, opened 2023-03-03
Convert FFI::DynamicLibrary to TypedData

larskanis:
And change a remaining use of TypedData object for completeness even if this code part is commented out.

Ref: #991


larskanis:
@casperisfine Thank you for all your work on ffi! This is the last class to be converted to TypedData. A good point to start introducing write barriers.

casperisfine:
Thanks for the merge and reviews!

I'll start adding write barrier and other bells and wistle next week!

PR #1001, opened 2023-03-06
Implement Write Barrier and dsize for FFI::FunctionType

casperisfine:

Write barrier protected objects are allowed to be promoted to the old generation, which means they only get marked on major GC.

The downside is that the RB_BJ_WRITE macro MUST be used to set references, otherwise the referenced object may be garbaged collected.

This commit also implement a `dsize` function so that these instance report a more relevant size in various memory profilers.

### DIFF
--- a/ext/ffi_c/FunctionInfo.c
+++ b/ext/ffi_c/FunctionInfo.c
@@ -53,16 +53,19 @@ static VALUE fntype_allocate(VALUE klass);
 static VALUE fntype_initialize(int argc, VALUE* argv, VALUE self);
 static void fntype_mark(void *);
 static void fntype_free(void *);
+static size_t fntype_memsize(const void *);
 
 const rb_data_type_t rbffi_fntype_data_type = { /* extern */
-  .wrap_struct_name = "FFI::FunctionType",
-  .function = {
-      .dmark = fntype_mark,
-      .dfree = fntype_free,
-      .dsize = NULL,
-  },
-  .parent = &rbffi_type_data_type,
-  .flags = RUBY_TYPED_FREE_IMMEDIATELY
+    .wrap_struct_name = "FFI::FunctionType",
+    .function = {
+        .dmark = fntype_mark,
+        .dfree = fntype_free,
+        .dsize = fntype_memsize,
+    },
+    .parent = &rbffi_type_data_type,
+    // IMPORTANT: WB_PROTECTED objects must only use the RB_OBJ_WRITE()
+    // macro to update VALUE references, as to trigger write barriers.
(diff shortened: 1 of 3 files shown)


ext/ffi_c/FunctionInfo.c
```diff
@@ -110,6 +113,23 @@ fntype_free(void *data)
     xfree(fnInfo);
 }
 
+static size_t
+fntype_memsize(const void *data)
```
casperisfine, line 117: I think I accounted for all held data, but worst case these function don't have to be 100% accurate.

spec/ffi/function_type_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    expect(function_type.result_type).to be == FFI::Type::Builtin::INT
+    expect(function_type.param_types).to be == [ FFI::Type::Builtin::CHAR, FFI::Type::Builtin::ULONG ]
+  end
+
+  it 'has a memsize function' do
+    base_size = ObjectSpace.memsize_of(Object.new)
+
+    function_type = FFI::FunctionType.new(:int, [])
+    size = ObjectSpace.memsize_of(function_type)
+    expect(size).to be > base_size
```
casperisfine, line 20: The size can vary on platforms etc, so I usually use this pattern to ensure the function report something, but not assert a precise size.

spec/ffi/function_type_spec.rb
```diff
@@ -0,0 +1,27 @@
+#
+# This file is part of ruby-ffi.
+# For licensing, see LICENSE.SPECS
+#
+
+require File.expand_path(File.join(File.dirname(__FILE__), "spec_helper"))
+
+describe FFI::FunctionType, skip: RUBY_PLATFORM == "java" do
```
casperisfine, line 8: The `java` ext doesn't define this class, I suppose that's why it wasn't tested. But I think these specs are worth it for the MRI implementation.

casperisfine:
Ok, it's now 🟢 , moving on to the next one.

ext/ffi_c/FunctionInfo.c
larskanis: A FunctionInfo can have a closurePool associated for all closures that are created in relation to this FunctionInfo. I'll add at least the management data of the ClosurePool to the size.

spec/ffi/function_type_spec.rb
larskanis: Yes, FunctionInfo is just an internal class not exposed through the API.

ext/ffi_c/FunctionInfo.c
casperisfine: Sounds good, I'll have a look at it.

We canalso report the size of the pool divided by the number of refs to it, I've done that in `msgpack-ruby`.

PR #1005, opened 2023-03-06
Implement Write Barrier and dsize for FFI::Type

casperisfine:

Write barrier protected objects are allowed to be promoted to the old generation,
which means they only get marked on major GC.

The downside is that the RB_BJ_WRITE macro MUST be used to set references,
otherwise the referenced object may be garbaged collected.

This commit also implement a `dsize` function so that these instance
report a more relevant size in various memory profilers.

While I was at it I removed the `strdup` of `BuiltinType.name`,
I don't see why we wouldn't directly point at static memory, that
saves copying type names, and makes freeing builtin types easier.

NB: it's based on https://github.com/ffi/ffi/pull/1001 to avoid conflicts.


casperisfine:
Rebased.

larskanis:
The `strdup` was there since the beginning of FFI::Type, but it can be removed.

PR #1010, opened 2023-03-06
Implement Write Barrier and dsize for FFI::Pointer

casperisfine:

As well as FFI::MemoryPointer and FFI::AbstractMemory

Write barrier protected objects are allowed to be promoted to the old generation, which means they only get marked on major GC.

The downside is that the RB_BJ_WRITE macro MUST be used to set references, otherwise the referenced object may be garbaged collected.

This commit also implement a dsize function so that these instance report a more relevant size in various memory profilers.

@larskanis I'm curious whether `FFI::AbstractMemory.new` is ever meant to be used? If I make it raise no test fail, and I really don't see what the usage would be. 

I'm tempted to just undefine the allocation function.

### DIFF
--- a/ext/ffi_c/AbstractMemory.c
+++ b/ext/ffi_c/AbstractMemory.c
@@ -55,6 +55,7 @@
 # define RB_OBJ_STRING(obj) StringValueCStr(obj)
 #endif
 
+static size_t memsize(const void *data);
 static inline char* memory_address(VALUE self);
 VALUE rbffi_AbstractMemoryClass = Qnil;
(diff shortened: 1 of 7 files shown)


casperisfine:
I added FFI::Buffer as well, as it's mostly the same code.

larskanis:
Probably no. I couldn't find any use of `FFI::AbstractMemory.new` by the github search. And I don't have an idea what a `FFI::AbstractMemory` object could be used for. I opened #1013, probably similar of what you though about.

PR #1013, opened 2023-03-06
Remove allocator of AbstractMemory

larskanis:
This disables AbstractMemory.new, which has no practical use.

As discussed in #1010


byroot (approved):
Yup exactly what I had in mind.

PR #1015, opened 2023-03-15
Update gen/Rakefile for Ruby 3 compat

matoro:
Ruby 3 changed the 'mode' argument from an option hash to keyword argument.  This change is backwards-compatible with Ruby 2.

See: https://github.com/ruby/fileutils/commit/482de6d397742526d1111576e2791f9b7051e3c0


larskanis:
Thank you!

PR #1018, opened 2023-04-11
Add types.conf for sw_64-linux

F16shen:


larskanis:
"sw_64" is Shen Wei 64-bit, based on Alpha? Is it used today?

F16shen:
Yes, it's still using

PR #1021, opened 2023-04-14
Make FFI classes GC.compact friendly

larskanis:
This allows to relocate all FFI objects.

Also remove marking of `StructLayout->cache_row` in favour of clearing it at `GC.compact` only.


larskanis:
Tests for this has been added in commit 6b1bfd123f172d6bc75042adc192a703bb4bdee6 .

PR #1025, opened 2023-04-18
Remove use of RARRAY_PTR

peterzhu2118:
RARRAY_PTR is bad for GC performance as it exposes the internal buffer of the array meaning that the array becomes write barrier unprotected.


larskanis:
Thank you, makes sense!

PR #1026, opened 2023-04-20
Windows - Update LoadLibrary Error Message to include Error Code

cfis:
On Windows FormatMessageA can return a NULL message - for example when trying to load a 32bit dll from a 64 bit program (garbage.so in the test suite). Since the passed in buffer contains junk data the resulting error message is made up of garbage characters which makes it impossible to know what the problem is.

This commit updates the message so that it always returns the error code (in this case 193) plus message, which makes it possible to debug the problem.


larskanis:
Looks good! Thank you!

cfis:
Oh you are right! I'll push another MR.

PR #1028, opened 2023-04-20
MSVC Test Support

cfis:
This MR includes several small changes that enable MSVC 2022 to compile the libtest library in the spec/fixtures directory. With these changes, all tests pass on MSVC.

Note that the compile.rb file on Windows only works with mingw, so it would have to change to support MSVC. I assume that isn't important,  but if we wanted to go down that route at some point we could either:

*  add a Makefile for nmake like mkmf does
*  or likely better, and way more modern, add a simple CMakelists.text file and use cmake

### DIFF
--- a/spec/ffi/fixtures/ClosureTest.c
+++ b/spec/ffi/fixtures/ClosureTest.c
@@ -16,10 +16,10 @@
 
 double testClosureVrDva(double d, ...) {
     va_list args;
-    double (*closure)(void);
+    typedef double (*closure_fun)(void);
 
     va_start(args, d);
-    closure = va_arg(args, double (*)(void));
+    closure_fun closure = va_arg(args, closure_fun);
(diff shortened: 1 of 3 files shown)


larskanis:
Although I have less experience with cmake, I prefer a cross platform solution over separate branches. The test environments can usually be changed easily, so that introducing cmake should be simple. In addition the GitHub CI has mswin binaries for testing. Feel free to open a PR.

larskanis:
@cfis I just added a CI run on Github Actions for ffi on MSVC / MSWIN. For now `libtest.dll` is built with GNU tools, but the rest is running on MSVC compiled binaries. CI looks like so: https://github.com/ffi/ffi/actions/runs/10220222201/job/28280151493

cfis:
Awesome - thanks!

PR #1035, opened 2023-06-25
Add support for aarch64-windows

ntkme:
This PR adds support for aarch64-windows.

Native build of MRI on aarch64-windows does not exist today. However, together with the following two PRs, I was able to run use ffi with JRuby on native aarch64-windows JDK.

- https://github.com/jnr/jnr-ffi/pull/330
- https://github.com/jnr/jffi/pull/144


headius:
Can we get a release out with these changes? I'd love to be able to ship JRuby 9.4.4.0 with Windows-ARM64 support, and we need to update the ffi gem to get these changes.

Issue #1039, opened 2023-07-27
Write RBS files


ParadoxV5:
Hi folks. I am a user of Ruby FFI.
I am interested in integrating it into my RBS workflow. Introduced with Ruby 3, [ruby/rbs](https://github.com/ruby/rbs/) is a type-checking language for Ruby.

What are your interests in RBS? Ideally, the Gem library would manage, update and ship the RBS of a Gem. I can help out with filling in the RBS documents or, in the meantime, upload a set of stubs to [ruby/gem_rbs_collection](https://github.com/ruby/gem_rbs_collection)

> Note that if you are the maintainer of the gem, you should consider moving RBS files to the gem's code repository and including them in the gem package instead.
> ⸺ [`ruby/gem_rbs_collection`, `docs/CONTRIBUTING.md`, § Become a code owner](https://github.com/ruby/gem_rbs_collection/blob/efc8580c38aa501593a8735ac6476cb42aaf6787/docs/CONTRIBUTING.md#become-a-code-owner)

larskanis:
I think we can integrate the RBS files into the gem package. However I don't have any experience with RBS. I only read about it several times. It would be great if you could create an initial pull request.

larskanis:
RBS API description will be part of ffi-1.17.0, currently available as prerelease ffi-1.17.0.rc2.

larskanis:
Thank you @ParadoxV5 for all your great work regarding RBS!

lloeki:
Thank you @ParadoxV5!

I've been meaning to do just that and upstream these, which we'll now be able to remove:

- https://github.com/DataDog/libddwaf-rb/blob/bf280637edac308a579399dd8e62db63d83a5949/vendor/rbs/ffi/0/ffi.rbs
- https://github.com/DataDog/dd-trace-rb/blob/f7a8fb5d597316c0ba274ee87a28948078f5a969/vendor/rbs/ffi/0/ffi.rbs

(... provided the small bits I've crudely written are implemented but from a cursory look it seems like they are)

PR #1042, opened 2023-08-11
RBS for key modules and classes (mainly user-facing ones)

ParadoxV5:
A step towards #1039.

Written against [v1.16.3 docs](https://rubydoc.info/gems/ffi/1.16.3/FFI). Maintainers’ assistance needed to cover undocumented C-implemented APIs.

Blockers regarding the new dependency `rbs ~> 3.0`:
* It is incompatible with EOL Ruby version 2.5 which FFI gem still supports.
* JRuby does not support RBS.

Both of which can be worked around by not declaring the not-so-dependent `rbs ~> 3.0`.
But that’s like writing Yardocs without declaring Yardoc as a dep.

#### This PR

* FFI (top-level)
* AbstractMemory
  * Buffer
  * Pointer
    * AutoPointer
    * Function
    * MemoryPointer
* DataConverter
* Enum
  * Bitmask
* Enums
* Library
* NativeType
* Struct
  * ManagedStruct
  * Union
* StructByReference
* Type
  * ArrayType
  * FunctionType
  * StructByValue
* VariadicInvoker

<details><summary>

#### What still needs to be done

</summary>

* ConstGenerator
* DynamicLibrary
* Generator
* IO
* LastError
* LibraryPath
* NotFoundError
* NullPointerError
* Platform
* PlatformError
* StrPtrConverter
* StructGenerator
* StructLayout
* StructLayoutBuilder
* TypesGenerator

and also, perhaps, tests (with [RBS](https://github.com/ruby/rbs/blob/master/docs/sigs.md#testing-signatures) or [Steep](https://github.com/soutaro/steep))

</details>


ParadoxV5:
[`Pointer#autorelease=`](https://rubydoc.info/gems/ffi/1.16.3/FFI/Pointer#autorelease=-instance_method) checks whether the argument is `true` or not. This means every truthy value except `true` is actually equivalent to `false` here. It should respect Ruby’s truthiness with `RTEST`.

larskanis:
@ParadoxV5 Thank you very much for all your work!

What do you think about merging this? Having an incomplete RBS API definition is at least a good starting point. Or can it be a problem in the future to release something incomplete?

ParadoxV5:
@larskanis 👋

> What do you think about merging this?

I was planning to do some more user-facing classes. I’ve been away over December focusing on [Advent of Code 2023](https://adventofcode.com/2023).

> Having an incomplete RBS API definition is at least a good starting point. Or can it be a problem in the future to release something incomplete?

Stubs should be forward-compatible.

---

A reminder that the CI fails because RBS 3.x is incompatible with Ruby 2.5 (which was EOL since a couple of years ago). RBS will have to downgrade if there’re no plans to raise the minimum Ruby version.

ParadoxV5:
For consistency, custom typedefs should also be a `{Symbol => Type}` hashmap, no `String`s. But
https://github.com/ffi/ffi/blob/1403e20b4877ff2ffd4d2b0425c7c43ed2a2c25c/ext/ffi_c/Type.c#L281

ParadoxV5:
Happy New Year @larskanis! This is all the class/method I’ll include for the first batch. Let me know if you prefer my commits squashed.

larskanis:
There was some additional discussion in https://github.com/ParadoxV5/ruby-ffi/pull/1 . These changes are now included in this PR.

It mainly adds the ability to check the RBS files against the FFI test suite. It works by the following command line:
```
RBS_TEST_TARGET='FFI::*' RUBYOPT='-rrbs/test/setup'  bundle exec rake spec
```

Issue #1050, opened 2023-09-25
Segfault in `async_cb_dispatcher_mark ` following an upgrade to `ffi 1.16.1`


casperisfine:
I haven't yet dug in the core dump, but opening an issue here for reference.

We saw a huge spike of 3k segfaults in just a dozen minutes after the upgrade:

```
[BUG] Segmentation fault at 0x0000000000000000
ruby 3.2.2 (2023-09-23 revision 6e984f9449) [x86_64-linux]

-- Threading information ---------------------------------------------------
Total ractor count: 1
Ruby thread count for this ractor: 18

-- C level backtrace information -------------------------------------------
/usr/local/ruby/bin/ruby(rb_vm_bugreport+0x1ef) [0x564558ad0eff] vm_dump.c:812
/usr/local/ruby/bin/ruby(rb_bug_for_fatal_signal+0xfc) [0x564558c5bb1c] error.c:1029
/usr/local/ruby/bin/ruby(sigsegv+0x4d) [0x564558a207cd] signal.c:964
/lib/x86_64-linux-gnu/libpthread.so.0(__restore_rt+0x0) [0x7fc42e372420]
/artifacts/bundle/ruby/3.2.0/gems/ffi-1.16.1/lib/ffi_c.so(async_cb_dispatcher_mark+0x4) [0x7fc425bfa334] Function.c:167
/usr/local/ruby/bin/ruby(ractor_local_storage_mark_i+0x18) [0x5645589ca338] ractor.c:3092
/usr/local/ruby/bin/ruby(rb_st_foreach+0x76) [0x564558a2db46] st.c:1580
/usr/local/ruby/bin/ruby(ractor_mark+0x13c) [0x5645589caefc] ractor.c:3107
/usr/local/ruby/bin/ruby(gc_mark_children+0x837) [0x5645588e7597] gc.c:7311
/usr/local/ruby/bin/ruby(gc_marks_rest+0x189) [0x5645588e8b09] gc.c:7431
/usr/local/ruby/bin/ruby(gc_start+0xf49) [0x5645588e9b29] gc.c:8718
/usr/local/ruby/bin/ruby(newobj_alloc+0x60e) [0x5645588eda3e] gc.c:2434
/usr/local/ruby/bin/ruby(rb_wb_protected_newobj_of+0x88) [0x5645588ee788] gc.c:2886
/usr/local/ruby/bin/ruby(ary_new+0xd0) [0x564558bc03d0] array.c:802
[...]

PR #1051, opened 2023-09-25
Handle null pointer in async_cb_dispatcher mark and free

casperisfine:
Fix: https://github.com/ffi/ffi/issues/1050

I don't fully understand that code, so my fix might be totally off. But according to production crashes, `ctx` might be NULL sometimes.

cc @larskanis


larskanis:
Thanks Jean for the fix! Will release ffi-1.16.2 with it.

Issue #923, opened 2021-12-03
Use LD_LIBRARY_PATH when looking up libraries


bouk:
`LD_LIBRARY_PATH` is a standard Linux method of looking up shared libraries, it is useful in some cases when using non-standard library paths—I use [Nix](https://nixos.org) for example which doesn't use the top-level library paths.

There's more details on LD_LIBRARY_PATH at https://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html

jaen:
@bouk are you sure it doesn't? Maybe you're not putting the library path where you think you are?

Apparently this part [here](https://github.com/ffi/ffi/blob/master/lib/ffi/library.rb#L129) is only a fallback to the `dl_open` call [here](https://github.com/ffi/ffi/blob/master/ext/ffi_c/DynamicLibrary.c#L119) which should take into account `LD_LIBRARYPATH` (see documentation [here](https://man7.org/linux/man-pages/man3/dlopen.3.html)).

I had a similar issue with `typhoeus` needing `libcurl` (hence my upvote), but when I wrapped `ruby` to put `curl` on `LD_LIBRARY_PATH` it just started working (it is a bit baroque because I need old Ruby, but should explain the general idea):
```nix
{ lib, pkgs-ruby_2_7_3, ... }:
  let 
    pkgs        = pkgs-ruby_2_7_3;
    ruby        = pkgs.ruby_2_7.withPackages (ps: with ps; [ 
                    byebug pry pkgs.solargraph pkgs.rufo rubocop
                  ]);
    libraryPath = lib.makeLibraryPath (with pkgs; [ curl ]);
  in
    pkgs.stdenv.mkDerivation {
      name = "ruby_2_7_3";

      src = ruby;

      dontUnpack = true;

      nativeBuildInputs = with pkgs; [ makeWrapper ];

      installPhase = ''
        mkdir -p "$out/bin"
        cp -r "$src" "$out"

        makeWrapper "$src/bin/ruby" $out/bin/ruby \
          --prefix LD_LIBRARY_PATH : ${libraryPath}
      '';
    }
    ```
    
    As such, I'm not quite sure if `ffi` needs to do anything more here?

larskanis:
ffi usually respects `LD_LIBRARY_PATH` when loading libraries.

jdelStrother:
[...]

PR #1059, opened 2023-11-08
allow SEARCH_PATH to be extended with LD_LIBRARY_PATH and DYLD_LIBRAR…

glenndavy:
I'm hoping this might help address issues like https://github.com/ffi/ffi/issues/923.

Using Nix (on darwin) I often hit issues (particularly with libraries like vips that require ffi) where glib etc can't be found. By setting up one of the standard paths like LD_LIBRARY path or DYLD_LIBRARY_PATH I _should_ be able to move past this, however FFI only looks i a few standard places for libraries.   This patch should provide an easy way past this.


eregon:
dlopen() should already consider those if you specify a library name, e.g. `m` and not `libm.so` or so.

PR #1061, opened 2023-11-13
Update `NativeType` doc

ParadoxV5:
* Add missing `BUFFER_INOUT` and “missing” `LONGDOUBLE`
* Remove mentions of (legacy types?) `CALLBACK` and `CHAR_ARRAY` – they are not found in 1.16.3 even though chances are that they *are* supported by every computer
* Update `ARRAY`, `FUNCTION` and `STRUCT` to CamelCase and separate these class aliases from the `Type::Builtin` list
  * Remove `MAPPED` – [`Mapped` has its own doc](https://rubydoc.info/gems/ffi/FFI/Type/Mapped)
* Clarify that, while all constants are exported to `Type`, the aliases aren’t found in `FFI` and `NativeType`
  * This could rather be an oversight in the `A` macro
* Fix typos
  * Avoid writing `(C)`, which instead renders to `©` ([rendered docs on `rubydoc.info`](https://rubydoc.info/gems/ffi/1.16.3/FFI/NativeType))

### DIFF
--- a/ext/ffi_c/Type.c
+++ b/ext/ffi_c/Type.c
@@ -333,10 +333,10 @@ rbffi_Type_Init(VALUE moduleFFI)
     classBuiltinType = rb_define_class_under(rbffi_TypeClass, "Builtin", rbffi_TypeClass);
     /*
      * Document-module: FFI::NativeType
-     * This module defines constants for native (C) types.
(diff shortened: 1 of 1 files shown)


larskanis:
Thank you, good catch!

PR #1075, opened 2024-01-25
Add Ruby 3.3 to CI

orgads:


orgads:
Thanks for the fast merge! When is the next release planned?

PR #1083, opened 2024-03-28
Update FFI::Pointer#== to not throw an exception

cfis:
If you call FFI::Pointer#== with an instance of an object that is not FFI::Pointer or descendants an exception is thrown:

> "INTERNAL ERROR!!! TypeError: wrong argument type Module (expected FFI::Pointer).

In my case, this then causes the RubyMine debugger (debase) to blow up, making debugging harder.

This PR updates FFI::Pointer#== to return false if a pointer is compared to a non-pointer object, which is the expected behavior.

### DIFF
--- a/ext/ffi_c/Pointer.c
+++ b/ext/ffi_c/Pointer.c
@@ -314,7 +314,12 @@ ptr_equals(VALUE self, VALUE other)
         return ptr->memory.address == NULL ? Qtrue : Qfalse;
     }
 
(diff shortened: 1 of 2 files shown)


eregon:
Could you add a test for this?

cfis:
Sure, added a few tests for Pointer#==

larskanis:
Thank you, looks good!

PR #1096, opened 2024-04-22
Add FFI::AbstractMemory#read_array_of_string

### LINKED ISSUE #1070: There’s an unexposed `AbstractMemory#read_array_of_string`
While reading https://github.com/ffi/ffi/pull/1064/commits/d5e7a0d, I noticed there’s an `AbstractMemory#read_array_of_string` that – isn’t accessible from Ruby (nor by Yardoc).

https://github.com/ffi/ffi/blob/1403e20b4877ff2ffd4d2b0425c7c43ed2a2c25c/ext/ffi_c/AbstractMemory.c#L484-L503

[P.S.] A reminder to double check if there’re other methods accidentally unexposed.

larskanis:
It was defined but not exposed to Ruby nor tested.

Fixes #1070

### DIFF
--- a/ext/ffi_c/AbstractMemory.c
+++ b/ext/ffi_c/AbstractMemory.c
@@ -1109,6 +1109,7 @@ rbffi_AbstractMemory_Init(VALUE moduleFFI)
     rb_define_method(classMemory, "read_bytes", memory_read_bytes, 1);
     rb_define_method(classMemory, "write_bytes", memory_write_bytes, -1);
     rb_define_method(classMemory, "get_array_of_string", memory_get_array_of_string, -1);
+    rb_define_method(classMemory, "read_array_of_string", memory_read_array_of_string, -1);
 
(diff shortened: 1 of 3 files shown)


larskanis:
@eregon This method is missing in Truffleruby. We could implement it here as a ruby method, but since JRuby already implements it natively, I tend to do it equally in all 3 implementations. Would you mind to add `FFI::AbstractMemory#read_array_of_string` to Truffleruby?

eregon:
Yeah, I will add it.
Since https://github.com/ffi/ffi/blob/master/lib/ffi/pointer.rb doesn't have `get_array_of_string`, I agree, better to keep these related methods together.

eregon:
Will be fixed in https://github.com/oracle/truffleruby/pull/3550

eregon:
@larskanis I restarted the truffleruby CI jobs, it's all green now, so I'll merge this.

eregon (approved):

PR #1097, opened 2024-04-22
Remove AutoPointer.new with block from documentation

### LINKED ISSUE #1071: Contrary to documentation, `AutoPointer#initialize` ignores block
```ruby
FFI::AutoPointer.new(FFI::Pointer::NULL) {} #=! no release method defined (RuntimeError)
```

larskanis:
It was never implemented since it doesn't work.

Fixes #1071


eregon:
It could work like `AutoPointer.new(ptr, &SomeClass.method(:release))`, but the `&` just seems extra so agreed.

ParadoxV5:
> `AutoPointer.new(ptr, &SomeClass.method(:release))`

When https://bugs.ruby-lang.org/issues/12125 passes, this could look like `AutoPointer.new(ptr, &SomeClass.:release)`

PR #1102, opened 2024-05-31
Install bigdecimal unconditional

larskanis:
.. since it otherwise disturbs parallel build with rake gem:native


lusinh:
@larskanis Stop this pull request. bigdecimal required ruby 3. It make a lot of repo with ruby 2 fail to bundle.

PR #1116, opened 2024-08-12
Support LD_LIBRARY_PATH on Linux

paholg:
Currently, we only check `LD_LIBRARY_PATH` on MacOs. I'm not sure why, but it means that we completely ignore it on Linux. This commit fixes that, so that we check it on all Unix systems.


larskanis:
LD_LIBRARY_PATH is already recognized by the Linux dynamic loader. What issue do you want to solve?

paholg:
Ah, that was not my experience, and this did fix it for me. Though now I can't reproduce the issue I was having.

I'll go ahead and close this and dig deeper if I run into an issue again. Thanks!

Issue #1114, opened 2024-08-09
Issue with post-fork callback execution (FFI Callback Dispatcher not restarted)


mensfeld:
Repro:

Run this docker container to get Kafka: https://github.com/karafka/karafka/blob/master/docker-compose.yml with `docker-compose up`

Save and run the below code. It will  install my [rdkafka](https://rubygems.org/gems/rdkafka) gem and run:

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'rdkafka'
end

require 'rdkafka'
require 'securerandom'

def create_topic(name = SecureRandom.uuid)
  Rdkafka::Config
    .new('bootstrap.servers': '127.0.0.1:9092')
    .admin
    .create_topic(name, 1, 1)
    .wait

  puts "[#{Process.pid}] Topic #{name} created!"
end

# This will work correctly and will create topic and not hang
create_topic

# Now we will fork ourselves few times
fork do
  # Expectation is for this to also print create but will hang on `#wait`
  # because the callback never kicks in
  create_topic
end

# Lets give it a moment
sleep(5)

# Now this will work
fork do
  # Now if I do the same but force-overwrite the already existing rdkafka
  # callback, all will work
  # Remove it not to get irrelevant warnings
  Rdkafka::Callbacks.send(:remove_const, :BackgroundEventCallbackFunction)
  Rdkafka::Callbacks::BackgroundEventCallbackFunction = FFI::Function.new(
      :void, [:pointer, :pointer, :pointer]
  ) do |client_ptr, event_ptr, opaque_ptr|
    Rdkafka::Callbacks::BackgroundEventCallback.call(client_ptr, event_ptr, opaque_ptr)
  end

  create_topic
end

sleep(5)
```

I was able to establish, that upon fork the `FFI Callback Dispatcher` thread is not started back.

This is how I define the callback function: https://github.com/karafka/rdkafka-ruby/blob/main/lib/rdkafka/callbacks.rb#L153

If that's me doing something wrong I would love to get info how to fix that.

KJTsanaktsidis:
I think we might need something like this? https://github.com/ffi/ffi/commit/62005f00e21c90eaba9657d388261774ce033bf6

If we agree I can polish that up and turn it into a PR next week.

mensfeld:
Yeah, we definitely need this otherwise, it will be dangling not only in rdkafka. Thank you :pray:

PR #1117, opened 2024-08-16
Restart async callback dispatcher thread after fork

KJTsanaktsidis:
If we don't do this, callbacks registered before fork will never actually fire after forking, and just hang forever.

We explicitly re-initialize the entire context (including re-initializing the mutexes and clearing out the list of callbacks), because:

* If a thread in the parent was in the process of calling a callback when we fork, the mutex would otherwise be "stuck on"
* I don't think we _want_ to actually execute any pending callbacks from other threads in the child process, so clearing the list makes sense anyway.

Fixes: https://github.com/ffi/ffi/issues/1114

### DIFF
--- a/ext/ffi_c/Function.c
+++ b/ext/ffi_c/Function.c
@@ -215,6 +215,52 @@ async_cb_dispatcher_set(struct async_cb_dispatcher *ctx)
     async_cb_dispatcher = ctx;
 }
 #endif
+
+static void
+async_cb_dispatcher_initialize(struct async_cb_dispatcher *ctx)
+{
+        ctx->async_cb_list = NULL;
+
+#if !defined(_WIN32)
+        /* n.b. we _used_ to try and destroy the mutex/cond before initializing here,
+         * but it's undefined what happens if you try and destory an unitialized cond.
+         * glibc in particular seems to wait for any concurrent waiters to finish before
+         * destroying a condvar, trying to destroy a condvar after fork that someone was
+         * waiting on pre-fork won't work. Just re-init he memory directly. */
+        pthread_mutex_init(&ctx->async_cb_mutex, NULL);
+        pthread_cond_init(&ctx->async_cb_cond, NULL);
+#else
+        InitializeCriticalSection(&ctx->async_cb_lock);
+        ctx->async_cb_cond = CreateEvent(NULL, FALSE, FALSE, NULL);
+#endif
+    ctx->thread = rb_thread_create(async_cb_event, ctx);
+
+    /* Name thread, for better debugging */
+    rb_funcall(ctx->thread, rb_intern("name="), 1, rb_str_new2("FFI Callback Dispatcher"));
+}
+
+static struct async_cb_dispatcher *
+async_cb_dispatcher_ensure_created(void)
+{
+    struct async_cb_dispatcher *ctx = async_cb_dispatcher_get();
+    if (ctx == NULL) {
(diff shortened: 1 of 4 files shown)


KJTsanaktsidis:
:( I accidently pushed a few versions of this PR which hang forever, and github seems to be waiting for all the previous enqueued jobs to time out (after six hours by default! :o) before running the tests on the latest version of this PR. So I guess I'm going to have to wait till tomorrow to get the test results here.

larskanis:
I canceled the CI jobs. Looks good now!

KJTsanaktsidis:
Thanks! Looks like I ticked all the old rubies the right way, I think

ext/ffi_c/Function.c
```diff
@@ -215,6 +215,52 @@ async_cb_dispatcher_set(struct async_cb_dispatcher *ctx)
...
+        ctx->async_cb_list = NULL;
+
+#if !defined(_WIN32)
+        /* n.b. we _used_ to try and destroy the mutex/cond before initializing here,
+         * but it's undefined what happens if you try and destory an unitialized cond.
+         * glibc in particular seems to wait for any concurrent waiters to finish before
+         * destroying a condvar, trying to destroy a condvar after fork that someone was
+         * waiting on pre-fork won't work. Just re-init he memory directly. */
+        pthread_mutex_init(&ctx->async_cb_mutex, NULL);
+        pthread_cond_init(&ctx->async_cb_cond, NULL);
```
ivoanjo, line 231: Minor: I wonder if, given the comment, it would be cleaner to use

```c
ctx->async_cb_mutex = PTHREAD_MUTEX_INITIALIZER;
ctx->async_cb_cond = PTHREAD_COND_INITIALIZER;
```

KJTsanaktsidis: I guess they're equivalent, yeah. I can change this if we prefer.

mensfeld (approved):

beauraF:
Hello 👋

Just to let you know that we have deploy this branch in our test environment and this fix a fork safety issue we had with one of our internal library that use FFI.

Thanks a lot @KJTsanaktsidis 🙏

mensfeld:
@beauraF you can use the karafka rdkafka fix meanwhile (rebuild the needed classes post-fork) to mitigate this.

FYI I am also waiting on this.

beauraF:
> without any other side effects as of today

I'd like to come back on this, one of our CI pipelines is blocked when we use this branch. Unfortunately I haven't had time to investigate further yet. It works when we use @mensfeld workaround

larskanis:
> I'd like to come back on this, one of our CI pipelines is blocked when we use this branch.

Thank you @KJTsanaktsidis for this great work! I think this can be merged, but the blocking CI should be investigated first. @beauraF It would be great if you can investigate it, or you can provide more details.

I didn't notice any side effects on my own, but I don't have any use cases with fork() other than the ffi specs.

KJTsanaktsidis:
Yeah @beauraF if you're able to provide some details, even if not a full reproduction, I can have a little stare at the code and think about what might be happening.

beauraF:
> @beauraF It would be great if you can investigate it, or you can provide more details.

> Yeah @beauraF if you're able to provide some details, even if not a full reproduction, I can have a little stare at the code and think about what might be happening.

So.. I just investigate, and the issue was fully on our side. Sorry about that..
I confirm this is fixing our issue. Thanks again 🙏

KJTsanaktsidis:
Great news, thanks for confirming!

PR #1130, opened 2024-12-04
Auto-convert Integer arguments to Pointer

danini-the-panini:
Reverts https://github.com/ffi/ffi/commit/59b48a924b27c1361ac2095a7f4a6668ed20eed5 and https://github.com/ffi/ffi/commit/5104827f3f8ae91dabba5df0b11600dd96d1834c

Fiddle supports auto-converting anything that can be coerced into an integer to Pointer, however this doesn't work in JRuby using the ffi_backend.

One solution to this is to add the integer auto-convert logic into the ffi_backend (https://github.com/ruby/fiddle/pull/162)

Alternatively, we can add integer auto-convert back into FFI both here and in JRuby (https://github.com/jruby/jruby/pull/8423)

The next question is, if we do this, should it only be for `Integer` (as originally implemented), or should it also allow any integer-coercible value (as implemented by Fiddle)

### DIFF
--- a/spec/ffi/pointer_spec.rb
+++ b/spec/ffi/pointer_spec.rb
@@ -61,7 +61,7 @@ describe "Pointer" do
     magic = 0x12345678
     memory.put_int32(0, magic)
     adr = memory.address
-    expect(PointerTestLib.ptr_ret_int32_t(adr, 0))
+    expect(PointerTestLib.ptr_ret_int32_t(adr, 0)).to eq(magic)
   end
 
   it "String can be used as a Pointer argument" do
--- a/spec/ffi/pointer_spec.rb
+++ b/spec/ffi/pointer_spec.rb
@@ -56,12 +56,12 @@ describe "Pointer" do
     expect(PointerTestLib.ptr_ret_int32_t(ptr, 0)).to eq(magic)
   end
 
-  it "Integer cannot be used as a Pointer argument" do
-    expect { PointerTestLib.ptr_ret_int32_t(0, 0) }.to raise_error(ArgumentError)
-  end
-
-  it "Bignum cannot be used as a Pointer argument" do
-    expect { PointerTestLib.ptr_ret_int32_t(0xfee1deadbeefcafebabe, 0) }.to raise_error(ArgumentError)
+  it "Integer can be used as a Pointer argument" do
+    memory = FFI::MemoryPointer.new :long_long
+    magic = 0x12345678
+    memory.put_int32(0, magic)
+    adr = memory.address
+    expect(PointerTestLib.ptr_ret_int32_t(adr, 0))
   end
 
   it "String can be used as a Pointer argument" do
--- a/ext/ffi_c/Call.c
+++ b/ext/ffi_c/Call.c
@@ -443,6 +443,10 @@ getPointer(VALUE value, int type)
 
         return StringValuePtr(value);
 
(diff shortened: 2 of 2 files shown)


spec/ffi/pointer_spec.rb
```diff
@@ -56,12 +56,12 @@ def to_ptr
...
-  end
-
-  it "Bignum cannot be used as a Pointer argument" do
-    expect { PointerTestLib.ptr_ret_int32_t(0xfee1deadbeefcafebabe, 0) }.to raise_error(ArgumentError)
+  it "Integer can be used as a Pointer argument" do
+    memory = FFI::MemoryPointer.new :long_long
+    magic = 0x12345678
+    memory.put_int32(0, magic)
+    adr = memory.address
+    expect(PointerTestLib.ptr_ret_int32_t(adr, 0))
```
eregon, line 64:
```suggestion
    expect(PointerTestLib.ptr_ret_int32_t(adr, 0)).to eq(magic)
```

eregon:
IMO I think we shouldn't treat plain integers as pointers, FFI in general is more "typed" and Fiddle is way less "typed", in fact it doesn't have helper methods to e.g. read byte/short/int or floats or doubles, contrary to FFI.
IOW, I think https://github.com/ruby/fiddle/pull/162 is the way to go, and I think FFI is best as it is about this behavior.

headius:
> IMO I think we shouldn't treat plain integers as pointers

I don't have a strong opinion here. I do like that FFI largely abstracts pointers behind a specific type, and no other values can be passed as pointers, but then we also want to support APIs like Fiddle that have no such restrictions.

Patching Fiddle in the short term is probably the best option, but I think we should leave the discussion open about allowing numeric pointer values to be passed into FFI functions directly. At the very least, this change would ease the hassle of working with native APIs that pass and return pointer values; we would not have to constantly re-wrap those values in Pointer objects, even if we did lose some type-safety around what constitutes a "pointer".

eregon:
> we would not have to constantly re-wrap those values in Pointer objects, even if we did lose some type-safety around what constitutes a "pointer".

Is that common? I don't recall seeing this much in FFI usages.
It sounds unlikely to me, either it should be declared as `something*` both as argument & return value and it just passes the Pointer object as-is in between, or it's declared as some kind of integer on both sides and it just works by passing the Integer.
It seems also pretty rare one would need to pass an Integer as a pointer value, where does this Integer come from?

larskanis:
I'm against accepting Integer as a Pointer input. I wrote a lot of different FFI based code and I used wrapping of integers by `Pointer.new(int)` only in the rare case when I wanted to trigger a segfault. In all other cases usually `MemoryPointer.new(...)` is used to get a valid memory pointer or `FFI::Pointer::NULL` to get a NULL pointer or `Pointer#get_pointer` to read a `Pointer` object from some memory. A `Pointer` object can even do a little bit of pointer arithmetic.

I know it is common in Fiddle to pass a `0` in case of a NULL pointer, but the separation of pointers and integers is IMHO a simple approach to give at least a little bit type safety and to avoid unnecessary confusion of the user by segfaults due to passing invalid parameters. Allowing Integers as Pointer input will IMHO increase the number of unwanted segfaults much more than the number of wanted use cases.

headius:
Seems like folks are leaning toward just making Fiddle do the extra work to wrap numeric pointers in the FFI backend.

eregon:
Let's close based on @larskanis's comment

PR #1142, opened 2025-02-17
Link to the classes and methods documentation

eregon:


eregon:
Mmh for some reason https://www.rubydoc.info/gems/ffi/FFI/Pointer is way worse than in an older version https://www.rubydoc.info/gems/ffi/1.16.3/FFI/Pointer : less methods, doesn't know it inherits from AbstractMemory.
In fact AbstractMemory is almost empty on https://www.rubydoc.info/gems/ffi/FFI/AbstractMemory but is better on https://www.rubydoc.info/gems/ffi/1.16.3/FFI/AbstractMemory

eregon:
1.71.0 seems also fine https://www.rubydoc.info/gems/ffi/1.17.0/FFI/Pointer so it seems only 1.17.1 is very incomplete.
Maybe some issue during docs generation or maybe due to actual changes in this repo?

larskanis:
Local generated `yard` documentation looks good, but https://www.rubydoc.info/gems/ffi/1.17.1/FFI/Pointer misses all C files. No idea why this happens. Even the [gem diff](https://my.diffend.io/gems/ffi/1.17.0/1.17.1) doesn't show me a good reason.

eregon:
> Local generated yard documentation looks good

Then maybe we should do something like https://github.com/ruby-concurrency/concurrent-ruby/blob/master/.github/workflows/docs.yml which automatically publishes the docs to GitHub pages.

eregon:
https://bugs.ruby-lang.org/issues/21350#note-1 has [a simple workflow](https://github.com/ruby/irb/blob/master/.github/workflows/gh-pages.yml) for hosting RDoc docs as GitHub Pages.
That seems a good solution.

eregon:
The problem seems to have resolved itself somehow, at least
https://www.rubydoc.info/gems/ffi/FFI/Pointer
and
https://www.rubydoc.info/gems/ffi/1.17.1/FFI/Pointer
looks as good as
https://www.rubydoc.info/gems/ffi/1.17.0/FFI/Pointer

So let's merge this.

PR #1149, opened 2025-05-08
Add spec for ptr.read_pointer.type_size

eregon:
Add spec for ptr.read_pointer.type_size
* See https://github.com/oracle/truffleruby/issues/3837

Add more specs for #type_size of pointers being read


eregon:
Should wait for https://github.com/oracle/truffleruby/pull/3851 to make the new specs pass on TruffleRuby.

PR #1158, opened 2025-08-21
Fix problematic type signatures

### LINKED ISSUE #1107: `sig` as packaged in 1.17.0 contain problematic RBS
We use typing heavily, type checked by Steep. We also use FFI as a dependency.

Previously we typed the minimum amount of FFI things and vendored our version of it with a plan to upstream, but we were ninja'd! See https://github.com/ffi/ffi/issues/1039#issuecomment-2147221540

Now, with `ffi` 1.17.0 packaging `sig` we'd be happy to use those, but there are some oddities that make Steep balk.

To wit:

```
sig/ffi/struct.rbs:23:29: [error] Type application of `::FFI::Type::Mapped` doesn't satisfy the constraints: ::FFI::Struct::ManagedStructConverter[::FFI::ManagedStruct[::FFI::AutoPointer, untyped], ::FFI::AutoPointer] <: ::FFI::_DataConverter[::FFI::Pointer, instance, untyped]
│ Diagnostic ID: RBS::UnsatisfiableTypeApplication
│
└     def self.auto_ptr: () -> Type::Mapped[
                               ~~~~~~~~~~~~~

sig/ffi/struct.rbs:5:15: [error] Type application of `::FFI::Type::Mapped` doesn't satisfy the constraints: ::FFI::StructByReference[::FFI::Struct[::FFI::AbstractMemory, untyped], ::FFI::AbstractMemory] <: ::FFI::_DataConverter[::FFI::AbstractMemory, instance, untyped]
│ Diagnostic ID: RBS::UnsatisfiableTypeApplication
│
└     type ptr = Type::Mapped[
                 ~~~~~~~~~~~~~

sig/ffi/auto_pointer.rbs:15:65: [error] Type application of `::FFI::AutoPointer::Releaser::_Proc` doesn't satisfy the constraints: self <: ::FFI::Pointer
│ Diagnostic ID: RBS::UnsatisfiableTypeApplication
│
└     def initialize: (Pointer pointer, Method | ^(self) -> void | Releaser::_Proc[self] callable) -> self
                                                                   ~~~~~~~~~~~~~~~~~~~~~

sig/ffi/library.rbs:36:45: [error] Type `::FFI::DataConverter` is generic but used as a non generic type
│ Diagnostic ID: RBS::InvalidTypeApplication
│
└     def typedef: [T < Type] (T old, Symbol | DataConverter add, ?untyped) -> T
                                               ~~~~~~~~~~~~~

Detected 4 problems from 3 files
```

These issues come from `ffi` itself, and can be reproduced with:

```
# add steep next to rbs
ruby -i -n -e "\$_ =~ /^(\s*gem 'rbs', '~> 3.0')/ and puts \"gem 'steep'\"; puts \$_" Gemfile

# install
bundle install

# set up steep
cat > Steepfile <<'STEEP'
target :lib do
  signature "sig"
  check "lib"
  ignore "lib" # don't check rb, as it fails currently
end
STEEP

# run check, which will only process rbs due to the ignore above
bundle exec steep check
```

y9v:
This PR fixes type signatures for `FFI::AutoPointer#initialize`, `FFI::StructByReference#native_type`, and `FFI::Struct#auto_ptr`.

There is also a commit adding `steep` dependency to `:type_check` bundle group and `Steepfile`. I can revert it if you don't want to keep it.

Should close #1107.


p-datadog:
@larskanis Could you run CI again on this PR? @y9v repaired the issues identified in the previous run.

PR #1182, opened 2026-03-25
Add a README section with differencies to fiddle gem [ci skip]

larskanis:


README.md
```diff
@@ -84,6 +84,39 @@ or from the git repository on github:
...
 * `--disable-libffi-alloc` : Force closure allocation by builtin method
 
+## Compare ffi to fiddle
+
+There's an alternative gem to call native functions called Fiddle: https://github.com/ruby/fiddle
+Both gems use [libffi](https://github.com/libffi/libffi) under the hood.
+
+These are the major differences between FFI and Fiddle:
+
+* Fiddle is delivered as a bundled gem as part of the ruby standard library.
```
eregon, line 94: A bundled gem isn't really standard library, for example it would need to be in the Gemfile to be used in a program using Bundler. So there difference there is quite small, it only matters when not using Bundler.

README.md
```diff
@@ -84,6 +84,39 @@ or from the git repository on github:
...
+* Fiddle is more lightweight than FFI.
```
eregon, line 95: I'd phrase it like it has a much smaller API but also much less features

README.md
```diff
@@ -84,6 +84,39 @@ or from the git repository on github:
...
+  It has some support for structs but not no other user defined types and type conversions.
+* Fiddle has some features, which are specific to CRuby (MRI).
+  It can read and write CRuby's VALUE type and handle IO types.
+* Fiddle is more low-level than FFI.
+  For instances often handles pointers as raw integer values.
+* Fiddle can not handle asynchronous callbacks from non-ruby threads.
+  This can be a blocker for some use cases.
```
eregon, line 102: I'd drop this sentence, it seems implied

README.md
```diff
@@ -84,6 +84,39 @@ or from the git repository on github:
...
+* Fiddle has direct support for [MemoryView](https://docs.ruby-lang.org/en/master/contributing/memory_view_md.html).
+
+... and on the other hand:
+
+* FFI must be installed as a gem, it is not bundled with CRuby (but with JRuby and Truffleruby).
```
eregon, line 107:
```suggestion
* FFI must be installed as a gem, it is not bundled with CRuby (but it is bundled with JRuby and Truffleruby).
```

README.md
```diff
@@ -84,6 +84,39 @@ or from the git repository on github:
...
+* FFI is more heavyweight than Fiddle.
```
eregon, line 108: I would say has more features or is more feature-rich.

eregon (commented):
Sounds good but I think the current comparison is rather unclear.
This how I would compare them:

* FFI has a idiomatic high-level Ruby API with all the features you would expect.
* Fiddle has a minimalist (I'd call it cryptic even TBH) low-level API (which is hard to use) and no conversion of data types, it is mostly used for debugging CRuby internals (and not for for wrapping any large native library).

The phrases in `()` are probably too much my opinion/not neutral enough, so if we drop them we get:

* FFI has a idiomatic high-level Ruby API with all the features you would expect.
* Fiddle has a minimalist low-level API and no conversion of data types, it is mostly used for very targeted usages and debugging CRuby internals.

`it is mostly used ...` might be arguable so I'd simplify to just this:

* FFI has a idiomatic high-level Ruby API with all the features you would expect.
* Fiddle has a minimalist low-level API and no conversion of data types.

eregon:
https://rubygems.org/gems/ffi vs https://rubygems.org/gems/fiddle is also pretty clear, `fiddle` has 16% of the downloads of `ffi`. `reverse_dependencies` are also telling.

eregon:
Another aspect is FFI is pretty well documented, Fiddle's documentation is near non-existent so it's very difficult to use even if one wanted to.

eregon:
Even the only code example on `ri Fiddle` is not portable and only works on Linux:
```
== Example

Here we will use Fiddle::Function to wrap {floor(3) from
libm}[http://linux.die.net/man/3/floor]

  require 'fiddle'

  libm = Fiddle.dlopen('/lib/libm.so.6')

  floor = Fiddle::Function.new(
    libm['floor'],
    [Fiddle::TYPE_DOUBLE],
    Fiddle::TYPE_DOUBLE
  )

  puts floor.call(3.14159) #=> 3.0
```

eregon:
Another way to see it is `Fiddle` is mostly a thin wrapper for libdl, i.e. `dlopen` (`Fiddle.dlopen`) and `dlsym` (`Fiddle::Handle.sym` and its alias `[]`). I guess it uses libffi only to be able to call such functions since libdl doesn't provide that functionality but otherwise doesn't to use much of `libffi`.

eregon:
I guess what I'm trying to say is people seem to have already figured out which is one is easier to use (FFI, by a very large margin) and comparing concisely is good but going into too much details is probably hiding the main points: FFI is high-level idiomatic easy-to-use well-documented many features, Fiddle is none of that.

README.md
```diff
@@ -84,6 +84,39 @@ or from the git repository on github:
...
+  It has user defined types and more type checking and type conversion capabilities.
+* FFI avoids extensions specific to the underlying ruby implementation.
+  It can't handle CRuby's VALUE or IO types.
+* FFI is more high-level than Fiddle.
+  It tries to avoid invalid memory accesses where possible.
+* FFI can handle asynchronous callbacks from non-ruby threads.
+  It spawns a ruby thread to process the callback in ruby.
+
+Both FFI and Fiddle be be used at the same time within one ruby application.
```
uvlad7, line 117:
```suggestion
Both FFI and Fiddle can be used at the same time within one ruby application.
```

PR #1189, opened 2026-04-14
Use libc allocator for Pointer and MemoryPointer-owned storage

rwstauner:
Pointer/MemoryPointer storage was previously allocated with xmalloc, so when native code calls libc free() we can hit allocator mismatch and abort on the latest development version of Ruby.

To make ownership-transfer workflows safe, this switches pointer-owned backing storage to malloc/free. This includes MemoryPointer and Pointer paths used by clone/dup, #free, and finalization.

The scope is still narrow to pointer-owned storage (not Buffer or other internal structures).

Example usage in the wild that can crash before this change:

1. Create C string pointer in sassc (via ffi)
   - SassC::Native.native_string
   - FFI::MemoryPointer.from_string(string) + autorelease = false
   - https://github.com/sass/sassc-ruby/blob/v2.4.0/lib/sassc/native.rb#L54-L58
2. Pass pointer into libsass data context
   - make_data_context calls sass_make_data_context
   - https://github.com/sass/sassc-ruby/blob/v2.4.0/lib/sassc/native/native_context_api.rb#L15-L18
3. Engine lifecycle
   - create context, compile, then ensure delete_data_context
   - https://github.com/sass/sassc-ruby/blob/v2.4.0/lib/sassc/engine.rb#L25-L26
   - https://github.com/sass/sassc-ruby/blob/v2.4.0/lib/sassc/engine.rb#L43
   - https://github.com/sass/sassc-ruby/blob/v2.4.0/lib/sassc/engine.rb#L63
4. ffi allocation side (from_string => xmalloc)
   - memptr_s_from_string constructs MemoryPointer
   - https://github.com/ffi/ffi/blob/v1.17.4/ext/ffi_c/MemoryPointer.c#L182-L188
   - constructor allocates backing storage with xmalloc
   - https://github.com/ffi/ffi/blob/v1.17.4/ext/ffi_c/MemoryPointer.c#L106-L117
5. libsass takes ownership and later frees with libc free
   - store incoming pointer: ctx->source_string = source_string
   - https://github.com/sass/libsass/blob/3.6.6/src/sass_context.cpp#L363-L376
   - free on teardown: free(ctx->source_string)
   - https://github.com/sass/libsass/blob/3.6.6/src/sass_context.cpp#L590-L596

### DIFF
--- a/ext/ffi_c/MemoryPointer.c
+++ b/ext/ffi_c/MemoryPointer.c
@@ -112,7 +112,11 @@ memptr_malloc(VALUE self, long size, long count, bool clear)
 
     msize = size * count;
 
-    p->storage = xmalloc(msize + 7);
+    /*
+     * MemoryPointer buffers may be handed off to native code that
+     * may later release them with libc free().
+     */
+    p->storage = malloc(msize + 7);
     if (p->storage == NULL) {
         rb_raise(rb_eNoMemError, "Failed to allocate memory size=%ld bytes", msize);
         return Qnil;
@@ -141,7 +145,8 @@ memptr_free(VALUE self)
 
     if (ptr->allocated) {
         if (ptr->storage != NULL) {
-            xfree(ptr->storage);
+            /* See memptr_malloc: storage is malloc()-backed. */
+            free(ptr->storage);
             ptr->storage = NULL;
         }
         ptr->allocated = false;
@@ -155,7 +160,8 @@ memptr_release(void *data)
 {
     Pointer *ptr = (Pointer *)data;
     if (ptr->autorelease && ptr->allocated && ptr->storage != NULL) {
-        xfree(ptr->storage);
+        /* See memptr_malloc: storage is malloc()-backed. */
+        free(ptr->storage);
         ptr->storage = NULL;
     }
     xfree(ptr);
--- a/ext/ffi_c/Pointer.c
+++ b/ext/ffi_c/Pointer.c
@@ -181,11 +181,13 @@ ptr_initialize_copy(VALUE self, VALUE other)
     }
 
     if (dst->storage != NULL) {
-        xfree(dst->storage);
+        /* Keep pair with malloc() used for pointer-owned storage. */
+        free(dst->storage);
         dst->storage = NULL;
     }
 
-    dst->storage = xmalloc(src->size + 7);
+    /* Pointer-owned storage is libc-backed for allocator symmetry. */
+    dst->storage = malloc(src->size + 7);
     if (dst->storage == NULL) {
         rb_raise(rb_eNoMemError, "failed to allocate memory size=%lu bytes", src->size);
         return Qnil;
@@ -406,7 +408,8 @@ ptr_free(VALUE self)
 
     if (ptr->allocated) {
         if (ptr->storage != NULL) {
-            xfree(ptr->storage);
+            /* Keep pair with malloc() used for pointer-owned storage. */
+            free(ptr->storage);
             ptr->storage = NULL;
         }
         ptr->allocated = false;
@@ -469,7 +472,8 @@ ptr_release(void *data)
 {
     Pointer *ptr = (Pointer *)data;
     if (ptr->autorelease && ptr->allocated && ptr->storage != NULL) {
-        xfree(ptr->storage);
+        /* Keep pair with malloc() used for pointer-owned storage. */
+        free(ptr->storage);
         ptr->storage = NULL;
     }
     xfree(ptr);
--- a/spec/ffi/dup_spec.rb
+++ b/spec/ffi/dup_spec.rb
@@ -34,6 +34,14 @@ describe "Pointer#dup" do
     # first char will be excised
     expect(p2.get_string(0)).to eq("est123")
   end
+
+  it "cloned pointer can be freed" do
+    p1 = FFI::MemoryPointer.new(:char, 1024)
+    p1.put_string(0, "test123")
+    p2 = p1[1].dup
+
+    expect { p2.free }.not_to raise_error
+  end
 end
 
 
--- a/spec/ffi/memorypointer_spec.rb
+++ b/spec/ffi/memorypointer_spec.rb
@@ -76,6 +76,36 @@ describe "MemoryPointer argument" do
     expect(p2.get_int(0)).not_to eql 0
   end
 end
+
+describe "MemoryPointer allocated memory ownership" do
+  module ForeignFree
+    extend FFI::Library
+    ffi_lib FFI::Platform::LIBC
+
+    attach_function :free, [:pointer], :void
+  end
+
+  it "can be released by libc free when created with from_string" do
+    ptr = MemoryPointer.from_string("ffi")
+    ptr.autorelease = false
+
+    expect { ForeignFree.free(ptr) }.not_to raise_error
+  end
+
+  it "can be freed via MemoryPointer#free" do
+    ptr = MemoryPointer.new(8)
+    expect { ptr.free }.not_to raise_error
+  end
+
+  it "can be duplicated and both instances can be freed" do
+    original = MemoryPointer.new(8)
+    copy = original.dup
+
+    expect { copy.free }.not_to raise_error
+    expect { original.free }.not_to raise_error
+  end
+end
+
 describe "MemoryPointer return value" do
   module Stdio
     extend FFI::Library


eregon:
> we can hit allocator mismatch and abort on the latest development version of Ruby.

How did this change in Ruby dev?

In general I think the rule for malloc/free is whoever (as in which library, or some set of functions in a library) allocates it should free it, it's never safe to free something you're handed that you didn't allocate, because you don't know what malloc it is, or it might even be static data which is unsafe to free.

sassc is also EOL and archived now so it's probably not a great example as it's effectively guaranteed to break anyway with enough time on latest Ruby.

I sympathize with the idea to just make it work, but the bug here is in sassc.

There are some drawbacks as this might imply slower performance notably I'm thinking of
https://github.com/truffleruby/truffleruby/blob/db85920d7dc78a66e302da8a3a141e410c8df9fb/src/main/ruby/truffleruby/core/truffle/ffi/pointer.rb#L256
https://github.com/truffleruby/truffleruby/blob/master/src/main/java/org/truffleruby/extra/ffi/Pointer.java#L63-L66
which wouldn't be able to use `UNSAFE.allocateMemory()` and need an actual call to malloc().

rwstauner:
The crash is coming from https://github.com/ruby/ruby/commit/91619f0230c0e5a95c796c1bd4f784c151e15614#diff-a52f658e599946fb24b955f80e3a707a471940dc4860489c9a104ccb10cf07f1R8273-R8277

It would indeed be ideal to get off of such an old extension (and I'm working on that), but it also seemed like the purpose of this class of objects was to be able to hand them off to another library and in those cases they may not be freed with xfree.

rwstauner:
I'm definitely relying on community discussion to determine if this is desired or not.

larskanis:
Thank you for the very detailed description of the problem! And sorry for my late response. Unfortunately this change is not desired. The described kind of use case is broken API design, which I don't want to support in FFI. Also the use of xmalloc/xfree has two additions to malloc/free which are desired for FFI:

* `xmalloc` triggers garbage collection, if not enough memory is available
* `xmalloc` counts the allocated memory to trigger the GC when certain threshold is reached

Nevertheless you can easily fix the issue by using the libc allocator to provide the memory storage like so:

```ruby
module C
  extend FFI::Library
  ffi_lib FFI::Library::LIBC

  attach_function :malloc, [:size_t], :pointer
  attach_function :free, [:pointer], :void
end

str = "abcdef"
m = C.malloc(str.bytesize+1)
po = FFI::Pointer.new(m)
po.write_string(str)
po.read_string  # => "abcdef" 
C.free(po)
```

jhawthorn:
> * `xmalloc` triggers garbage collection, if not enough memory is available

This effectively never happens on production systems because of overcommit (well on Linux/FreeBSD at least). If that's the concern we could safely replace `malloc` with a function doing: `malloc` -> `rb_gc` -> `malloc (retry)`. But I think it's probably not worth it.

> * `xmalloc` counts the allocated memory to trigger the GC when certain threshold is reached

That's a noble goal, but the possibility of mismatching a xmalloc/free or vice versa does much more harm than good. CRuby clamps, resets, and avoids integer underflow (used to be a bug 😬) because this misbehaviour exists, but it's harmful and IMO should warn (That's also partly why it errors in debug now).

---

I'm pretty confident that these buffers should be `malloc`/`free`'d or this is an error in FFI's design.

It is and has been a bug forever to pair a xmalloc/free or malloc/xfree (because it messes up internal accounting). There has been a compile option that's existed for a long time (Ruby 1.9?), `CALC_EXACT_MALLOC_SIZE`, which would cause a crash if you did this. In Ruby 4.1dev this is enabled by default in debug builds (it's helpful for finding bugs like these and similar).

Libraries FFI uses are going to be calling malloc/free. Not xmalloc/xfree unless they are a Ruby C extension (in which case why is FFI here).

To me, "Autorelease" is the smoking gun.

> By default a pointer object frees its content when it's garbage collected. Therefore it's usually not necessary to call [#free](https://www.rubydoc.info/gems/ffi/FFI/Pointer#free-instance_method) explicit. This behaviour may be changed with [#autorelease=](https://www.rubydoc.info/gems/ffi/FFI/Pointer#autorelease=-instance_method) method. If it's set to `false`, the memory isn't freed by the garbage collector, but stays valid until `free()` is called on C level or when the process terminates.

I think that spells out very explicitly that the user is expected _sometimes_ to take this memory and call `free()` (not `xfree()`) on it, while setting `autorelease=false`. Again, it would be strange for library code to call `xfree`. So either this should use `malloc`/`free` or **autorelease should not be assignable** - A Pointer initialized with existing memory MUST be freed manually with a function, also `#free` should error and refuse to work - and memory allocated by FFI for the Pointer MUST be freed by `Pointer#free` or GC.

From the code @eregon linked it sounds like truffleruby is already making the assumption that `autorelease` is read-only by not using malloc? If so I'm surprised that doesn't cause issues.

larskanis:
> To me, "Autorelease" is the smoking gun.

Oh, I missed that point! Then it makes sense to change to malloc/free.

> * xmalloc counts the allocated memory to trigger the GC when certain threshold is reached

I still remember about the issue with malloc'd data in ruby-pg. It was described here: https://samsaffron.com/archive/2018/06/13/ruby-x27-s-external-malloc-problem

I fixed this issue per `rb_gc_adjust_memory_usage()` and I think we can use it in FFI as well. When the memory is allocated, then the malloc'd size might be added and when `#free` is called or `#autorelease=false` is called, it is decreased. So the common case of `#autorelease=true` is still noticed by the GC.

eregon:
I'm not following the "smoking gun" reasoning here, could you clarify it with concrete example(s)?
So far I'm thinking it's mostly the docs are wrong.

Either the pointer is allocated by FFI (i.e. `FFI::MemoryPointer`, with `xmalloc()`) and then if auto-release==true it is freed by Ruby GC and `xfree()`.
If auto-release==false, then `FFI::MemoryPointer#free` should be called, which is still `xfree()` and matching the xmalloc() (and that's where the docs are misleading).

If it's FFI::Pointer by default there is no auto-release by default, so users of that are responsible to both allocate and free and FFI is not involved in either.
The problem is then if calling `FFI::Pointer#free`, I agree that should be an error for non-FFI::MemoryPointer (i.e. not allocated by FFI).
It's probably also not a good idea to allow setting auto-release to true for FFI::Pointer, because we don't know what allocated it (maybe it was jemalloc() without LD_PRELOAD and then using free() would still be wrong).

Alternatively, `FFI::Pointer#free` should do nothing (or maybe just warn) when the pointer was not allocated by FFI.

---

Separately I wonder if it's possible for `malloc` in one dynamic library to refer to a different malloc in another library.
It would be trivially the case if a malloc implementation is included in either library, but that's probably rare.
But what if e.g library A depends on say glibc and so uses glibc malloc but library B depends on e.g. jemalloc and resolves malloc to jemalloc's version?
If any of these are possible then in general it's not possible for FFI to match what the library used.
And the only real solution is: whoever allocates should also free, because only they know how it was allocated.

We can also take the example of allocations from a pool, clearly only the code that allocated it knows how to free it, and everything else is incorrect and would probably crash.

eregon:
There is BTW `FFI::AutoPointer` to pass in a free function as a Proc.

rwstauner:
For reference this snippet appears to work as a monkey-patch for [sassc](https://github.com/sass/sassc-ruby/blob/v2.4.0/lib/sassc/native.rb#L54-L58).

Amusingly I got another allocator mismatch issue linking to `LIBC` in an environment that loads jemalloc because sassc ends up going through `je_free_default` and it segfaults.
Thankfully switching to `CURRENT_PROCESS` seems to work.

```ruby
module GlobalAllocator
  extend FFI::Library
  # FFI::CURRENT_PROCESS resolves symbols from the process-global namespace
  # (RTLD_DEFAULT), honouring any allocator interposition (jemalloc).
  ffi_lib FFI::CURRENT_PROCESS
  attach_function :malloc, [:size_t], :pointer
end

def native_string(string)
  bytes = string.to_s
  pointer = GlobalAllocator.malloc(bytes.bytesize + 1)
  Kernel.raise(NoMemoryError, "SassC::Native.native_string: malloc failed") if pointer.null?

  pointer.put_string(0, bytes)
  pointer
end
```

jhawthorn:
> If auto-release==false, then `FFI::MemoryPointer#free` should be called, which is still `xfree()` and matching the xmalloc() (and that's where the docs are misleading).

If so, why have `auto_release`. If you call `FFI::MemoryPointer#free` it won't be free'd on GC so that will be sufficient. That would mean the only feature `auto_release = false` provides is to let you leak memory. That doesn't make any sense.

> Separately I wonder if it's possible for `malloc` in one dynamic library to refer to a different malloc in another library.
> It would be trivially the case if a malloc implementation is included in either library, but that's probably rare.
> But what if e.g library A depends on say glibc and so uses glibc malloc but library B depends on e.g. jemalloc and resolves malloc to jemalloc's version?

Sure, that could happen in the case that a specific library that chose to use it's own allocator (EDIT: also maybe windows has more problems with this?). It's _extremely_ common though to be returned `malloc`'d memory with the expectation that the caller `free`s it with `free()`. For one of many examples, [`realpath(3)`](https://man7.org/linux/man-pages/man3/realpath.3.html) - "The caller should deallocate this buffer using [free(3)](https://man7.org/linux/man-pages/man3/free.3.html)."

eregon:
> If so, why have `auto_release`. If you call `FFI::MemoryPointer#free` it won't be free'd on GC so that will be sufficient. That would mean the only feature `auto_release = false` provides is to let you leak memory. That doesn't make any sense.

The default is `autorelease = true`, but it's useful to have it false, because that way:
* `FFI::MemoryPointer.new(size)` is still a convenient way to allocate native memory
* An explicit `.free` is less work for the GC, and removes the overhead of a finalizer
* An explicit `.free` is immediate and doesn't delay freeing the memory until the next (major I guess) GC
* In fact this is used quite a bit in TruffleRuby to implement the socket stdlib, etc, and it makes sense for many gems using FFI too.

BTW `FFI::MemoryPointer.new` takes a block and it's more efficient by not needing a finalizer. It's not always possible to have the logic in the block though.

I guess what you are saying is if one wants to `.free` manually they could just leave the default `autorelease = true` and it still works. True, though it keeps the memory footprint overhead of the finalizer and actually I'd like to have a way to create a FFI::MemoryPointer with `autorelease = false` at creation time to avoid finalizer overhead.

> It's _extremely_ common though to be returned `malloc`'d memory with the expectation that the caller `free`s it with `free()`.

Actually I think it's only a few libc functions which do that, realpath(), strdup() and a few others.
The most common case by far is taking an existing buffer, e.g. read/write/sprintf/....
Even realpath() has the option to take an existing buffer.
Also e.g. asprintf() which is sprintf() but allocates using malloc() is not even standard but a `_GNU_SOURCE` extension.

I think all of these are safe if that function and free() are used in the same compilation unit or in the same library (and let's consider an executable as a library too to use a common terminology).
But they are unsafe when malloc() happens in one library and free() happens in another. This is the case here, nothing guarantees malloc() as seen by libfoo() is the same as free() seen by libruby(). It would be trivially wrong if e.g. libfoo() depend on jemalloc or so.
So that's why I think we should not change this, it's inherently unsafe and I see it as a feature that we now detect more easily incorrect usages.

eregon:
> Amusingly I got another allocator mismatch issue linking to `LIBC` in an environment that loads jemalloc because sassc ends up going through `je_free_default` and it segfaults.

Exactly, that's why the only safe way is if both the malloc and free functions come from the same compilation unit or library.

rwstauner:
I found another instance of this issue in the `rdkafka`, same setup as `sassc` (use MemoryPointer with autorelease to hand ownership to another library that then tries to free it and we get an allocator mismatch):

https://github.com/karafka/rdkafka-ruby/pull/969

